PR #4, opened 2023-03-08
design tokens

piotrekzawada:


insanepl (commented):
There's PR with pnpm and storybook you can safely remove [package-lock.json](https://github.com/elpassion/flounder-2/pull/4/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519) file.

packages/design-system/src/components/Button/index.tsx
```diff
@@ -1,5 +1,24 @@
 import classNames from "classnames";
-import type { ButtonProps } from "./Button.interface";
+
+import { IconTypes } from "utils/iconType";
+
+type VariantType =
```
insanepl, line 5: We moved most of the interfaces to type files now, let's remember to refactor those types in new files and remove this.

packages/design-system/src/components/InlineMessage/index.tsx
```diff
@@ -1,6 +1,13 @@
 import classNames from "classnames";
-import type { InlineMessageProps } from "./InlineMessage.interface";
 
+import { IconTypes } from "utils/iconType";
+
+export interface InlineMessageProps {
```
insanepl, line 5: Same as above

packages/design-system/src/tokens/tokens.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface TokenValues {
```
insanepl, line 1: Maybe we can move this to /utils and then export as type with npm library

packages/design-system/tailwind.config.js
```diff
@@ -1,4 +1,12 @@
 /** @type {import('tailwindcss').Config} */
+const tailwindConfigHelper = require("./src/tokens/helpers");
```
insanepl, line 2: I understand this is default, but that's the step we need to do after installing to customize right, we can add this helpers funtions to our npm lib and export, then we just add usagage examples to README, or you have any other idea? Because I dont think we're gonna exporting taiwind config from npm lib ultimately

packages/design-system/src/components/Button/index.tsx
piotrekzawada: todo added

packages/design-system/src/components/InlineMessage/index.tsx
piotrekzawada: todo added

packages/design-system/src/tokens/tokens.interface.ts
piotrekzawada: this will be removed

packages/design-system/tailwind.config.js
piotrekzawada: I don't think either. It's working example not a final version I guess

PR #6, opened 2023-03-08
Fix storybook and pnpm issue / prepare modules / cleanup old repo files

insanepl:


pauzim (approved):

packages/design-system/src/components/Items/index.tsx
```diff
@@ -6,7 +6,7 @@ export const Container: React.FC<ContainerProps> = ({
   className,
 }) => {
   //TIP:classes to use for active state
-  const activeStyle = "bg-neutral-200 border-neutral-200";
+  // const activeStyle = "bg-neutral-200 border-neutral-200";
```
piotrekzawada, line 9: you sure about this?

insanepl: I commented unused variables so we weren't using it anyway

PR #13, opened 2023-03-10
ELPDES-123 - icon refactor

piotrekzawada:


packages/design-system/src/components/Icon/Icon.interface.ts
```diff
@@ -1,9 +1,8 @@
 import type { HTMLAttributes } from "react";
 import type { IconTypes } from "../../utils/iconType";
 
-export interface IconProps
-  extends HTMLAttributes<HTMLSpanElement> {
-  icon: IconTypes;
+export interface IconProps extends HTMLAttributes<HTMLSpanElement> {
+  icon: string;
```
insanepl, line 5: I would leave it as it was and add additional props iconName to keep compatibility in old components, and when we refactor all to use names then we can remove it completely

insanepl: https://github.com/elpassion/flounder-2/pull/16/files#diff-c8d10a97f3a5722442b0301684cc5bd45ffa6f951d4583ba9e5e4619bc82dcafR20

packages/design-system/src/stories/atoms/Icon.stories.tsx
```diff
@@ -1,20 +1,39 @@
 import React from "react";
```
insanepl, line 1: As you can see I renamed Icon to Icons.stories so Icon will be for actual icon component, make sure to rename it too in your to avoid conflicts when merging

insanepl (changes requested):

packages/design-system/src/stories/atoms/Icon.stories.tsx
```diff
@@ -1,15 +1,11 @@
...
   return (
     <>
-      ADD Icon Component Documentation
-      <br />
-      How to use with our font entity icons
-      <br />
-      How to use with customIcon ReactNode
-      <br />
-      Aailable Props
+      <IconComponent iconName="wifi" size="lg" />
```
insanepl, line 8: Maybe we cna prepare array with like few names to choose with and test icon looks?

PR #16, opened 2023-03-12
Feat/update npm package

insanepl:


packages/design-system/src/utils/EasyImage.tsx
```diff
@@ -0,0 +1,22 @@
+import classNames from "classnames";
+import type { ImgHTMLAttributes } from "react";
+
+interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {
+  src: string;
+  objectFitCover?: boolean;
+}
+export const EasyImage: React.FC<ImageProps> = ({
+  src,
+  alt = "User avatar",
```
piotrekzawada, line 10: let's not have any default value here - this won't be able to spot A11Y errors with same alt for every image with `alt` prop

packages/design-system/src/utils/EasyImage.tsx
```diff
@@ -0,0 +1,22 @@
+import classNames from "classnames";
+import type { ImgHTMLAttributes } from "react";
+
+interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {
+  src: string;
+  objectFitCover?: boolean;
+}
+export const EasyImage: React.FC<ImageProps> = ({
```
piotrekzawada, line 8: there is nothing easy about this image, what about `ImageTag` or sth more descriptive?

packages/design-system/src/components/Avatar/index.tsx
```diff
@@ -86,7 +82,7 @@ export const Avatar: React.FC<AvatarProps> = ({
     <div className="flex items-center">
       <AvatarComponents.Container size={size} {...props}>
         {src ? (
-          <AvatarComponents.Image src={src} />
+          <EasyImage src={src} className="w-full" objectFitCover />
```
piotrekzawada, line 85: let's not remove dot notation component. Instead removing `Image` let's replace the content with this component.

packages/design-system/src/components/ButtonGroup/ButtonGroup.interface.ts
```diff
@@ -1,7 +1,9 @@
+import type { ReactNode } from "react";
 import type { ButtonProps } from "../Button/Button.interface";
 
 export interface ButtonGroupProps extends Pick<ButtonProps, "size"> {
   buttons: Omit<ButtonProps, "size" | "variant">[];
   variant: "primary" | "outlined" | "ghost";
   style: "default" | "withIcon" | "withText";
+  withIcon?: ReactNode;
```
piotrekzawada, line 8: i don't like props starting with `with` or `is` - there is an if hidden with this kind of values. I'd rather ranem it from `withIcon`, which indicates boolean, to simply `Icon` which is more ReactNode

packages/design-system/src/components/Icon/Icon.interface.ts
```diff
@@ -1,9 +1,10 @@
-import type { HTMLAttributes } from "react";
+import type { HTMLAttributes, ReactNode } from "react";
 import type { IconTypes } from "../../utils/iconType";
 
-export interface IconProps
-  extends HTMLAttributes<HTMLSpanElement> {
-  icon: IconTypes;
+export interface IconProps extends HTMLAttributes<HTMLSpanElement> {
+  icon?: IconTypes;
+  customIcon?: ReactNode;
```
piotrekzawada, line 6: shouldn't we be using `TIcon` for icons?

packages/design-system/src/components/Search/index.tsx
```diff
@@ -35,33 +36,27 @@ const Search: React.FC<SearchProps> = ({
     <div
       className={classNames(
         "flex items-end gap-2",
-        // @TODO: Add special sizes to config and ditch JIT to keep design system consistent embeded into config
-        variant === "inline" && "!w-[100px]"
+        variant === "inline" && "min-w-min max-w-fit"
```
piotrekzawada, line 39: there is at least 4 repetition of this if. Can you refactor it a bit?

packages/design-system/src/svgs/HomeSvg.tsx
```diff
@@ -0,0 +1,20 @@
...
+      fill="none"
+      viewBox="0 0 24 24"
+      className={className}
+    >
+      <path
+        d="M9 22V12H15V22M3 9L12 2L21 9V20C21 20.5304 20.7893 21.0391 20.4142 21.4142C20.0391 21.7893 19.5304 22 19 22H5C4.46957 22 3.96086 21.7893 3.58579 21.4142C3.21071 21.0391 3 20.5304 3 20V9Z"
+        stroke="currentColor"
+        stroke-width="1.4"
+        stroke-linecap="round"
+        stroke-linejoin="round"
```
piotrekzawada, line 14: react won't be happy here. It should be `strokeWidth` etc.

piotrekzawada:
@insanepl let's make sure that all icon props uses TIcon and SVG prop are in correct format: `stroke-width` -> `strokeWidth` etc.

PR #24, opened 2023-03-14
Update package bundle strategy and solve minor problems

insanepl:


packages/design-system/src/helpers/icons/iconNames.ts
```diff
@@ -0,0 +1,291 @@
+export type IconNames =
```
piotrekzawada, line 1: i don't think it's good to keep names as fixed values - designers are able to change the file

PR #25, opened 2023-03-14
Enable avatar inside `Chips` component

pauzim:
Fix missing avatar option in storybook


packages/design-system/src/stories/utils.tsx
```diff
@@ -35,6 +36,16 @@ const rectangleImgIcon = (
   <FeaturedTextImage src={getImageUrl("/rectangle.png")} alt="rectangle" />
 );
 
+const peachRoundedAvatarImg = {
+  ...(
+    <ImageTag
```
insanepl, line 41: is Avatar component with size variant "xxs" too big?

insanepl: Also I see on Figma chips is always rounded, so maybe we shoudl code rounded variant inside Chips and just overflow-hidden on parent will take care of that

insanepl (changes requested):

packages/design-system/src/stories/utils.tsx
pauzim: I did this but with these styles icon doesn't fit and is clipped. Last time we discussed to get rid of variants but maybe we should use them and in this scenario icon won't be affected and only variant `avatar` will be rounded

piotrekzawada (approved):
LGTM

PR #31, opened 2023-03-16
Fix components with input

mzyznowska:
This PR contains search improvement [ELPDES-124](https://elpassion.atlassian.net/browse/ELPDES-124?atlOrigin=eyJpIjoiYTRkMGQyNjE2ZTJjNGFiMmJlMDMwN2Y1M2JmZmRkMmYiLCJwIjoiaiJ9)

[ELPDES-124]: https://elpassion.atlassian.net/browse/ELPDES-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/design-system/src/svgs/XCircleSvg.tsx
```diff
@@ -0,0 +1,20 @@
...
+        stroke="currentColor"
+        strokeWidth="1.4"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
+};
+
+export default UserSvg;
```
insanepl, line 20:
```suggestion
export default XCircleSvg;
```

insanepl: Also I don't think it's necessary I just checked by paths, we already got this one https://github.com/elpassion/flounder-2/blob/main/packages/design-system/src/svgs/CircularCloseSvg.tsx

packages/design-system/src/components/InputAutocomplete/index.tsx
```diff
@@ -81,28 +81,37 @@ export const InputAutocomplete: FC<InputAutocompleteProps> = ({
...
-  const isTyping = !!inputValue;
+  const inputRef = useRef<HTMLInputElement>(null);
+  const onChange = () => setInputValue(inputRef?.current?.value || "");
+  const isTyping = inputValue;
+
+  const clearInput = () => {
+    if (inputRef && inputRef?.current) {
+      inputRef.current.value = "";
+    }
+    setInputValue("");
```
insanepl, line 92: Why we just don't add change handler so we cna pass value and handler to input form? Then we dont need to pass current.value and setInputValue, seems like overcomplication to me

insanepl (changes requested):

packages/design-system/src/components/InputAutocomplete/index.tsx
mzyznowska: I removed `setInputValue("")` in this function(`clearInput`) because I forgot to do it but I need inputValue to pass it to dropdown. I don't know what solution do you propose.

insanepl: I mean that maybe we can pass onChange and value optionally as most libs do it, and we control value from top where component is mounted by wiring onChange to some setState or react hook form then value will come from controller

piotrekzawada (approved):
LGTM 🙏

insanepl (approved):

PR #37, opened 2023-03-20
ELPDES- 128:  `Legend` component 

pauzim:


packages/design-system/src/components/Legend/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <div
+        className={classNames(
+          "aspect-square w-4",
+          color,
+          labelShapeVariants[shape]
+        )}
+      />
+      <div>
+        <span>{label}</span>
+        {displayValues && <span className="pl-1">{`(${value})`}</span>}
```
piotrekzawada, line 55: let's get rid of this boolean. We can simply do sth like:
```suggestion
        <span className="pl-1 empty:hidden" >{`(${value})`}</span>
```

packages/design-system/src/components/Legend/Legend.interface.tsx
```diff
@@ -0,0 +1,20 @@
+export interface LegendProps {
+  entries: ILegendEntry[];
+  shape?: TLegendShape;
+  background?: string;
+  horizontal?: boolean;
```
piotrekzawada, line 5: I don't think passing a boolean is a good option. What about sth like in divider? https://github.com/elpassion/flounder-2/blob/main/packages/design-system/src/components/Divider/Divider.interface.ts#L4

piotrekzawada (approved):

PR #45, opened 2023-03-24
ELPDES-153 - rating

mzyznowska:


packages/design-system/src/components/Rating/index.tsx
```diff
@@ -0,0 +1,125 @@
+import { FC, useState } from "react";
+import classNames from "classnames";
+
+import StarSvg from "@svgs/StarSvg";
+import { TComponentIcon } from "@helpers/types";
+import HeartSvg from "@svgs/HeartSvg";
+
+export interface RatingProps {
```
pauzim, line 8: We can export this interface to the separate `.interface` file

packages/design-system/src/components/Rating/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <div
+        className={classNames(
+          "flex w-fit gap-1",
+          ratingsAlignVariantsStyle[ratingsAlign],
+          { "items-center": align === "center" }
+        )}
+      >
+        <p className="text-base text-black empty:hidden">{rate}</p>
+        <div className="flex gap-2 text-neutral-100">
+          {[1, 2, 3, 4, 5].map((index) => (
```
pauzim, line 97: 1. I think we should be flexible as much as possible and enable passing some component prop defining max rate (eg. `maxRate` or sth else more descriptive). We can assume `5` as default
2. Let's define some variable for ratings array and build it dynamically based on `maxRate`
suggestion: ```const rates=[...Array(maxRate).keys()].map(el=> el+1 )``` and map this array to display ⭐️

packages/design-system/src/stories/atoms/Rating.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return <RatingComponent {...props} rate={newRate} onClick={setNewRate} />;
+};
+
+export default {
+  title: "🟠 Atoms/Rating",
+  component: Rating,
+  argTypes: {
+    title: { description: "string" },
+    supportingText: { description: "string" },
+    rate: {
```
pauzim, line 23: Not sure if we should define this control as `number` because storybook treats this as required value in that scenario? 🤔

packages/design-system/src/svgs/SmileSvg.tsx
```diff
@@ -0,0 +1,25 @@
...
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <g clip-path="url(#clip0_4495_66434)">
+        <path
+          d="M6.66667 11.6667C6.66667 11.6667 7.91667 13.3333 10 13.3333C12.0833 13.3333 13.3333 11.6667 13.3333 11.6667M7.50001 7.5H7.50834M12.5 7.5H12.5083M18.3333 10C18.3333 14.6024 14.6024 18.3333 10 18.3333C5.39763 18.3333 1.66667 14.6024 1.66667 10C1.66667 5.39763 5.39763 1.66667 10 1.66667C14.6024 1.66667 18.3333 5.39763 18.3333 10Z"
+          stroke="currentColor"
+          stroke-width="1.4"
+          stroke-linecap="round"
+          stroke-linejoin="round"
```
pauzim, line 18:
```suggestion
          strokeWidth="1.4"
          strokeLinecap="round"
          strokeLinejoin="round"
```

packages/design-system/src/components/Rating/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+        {title}
+      </p>
+      <div
+        className={classNames(
+          "flex w-fit gap-1",
+          ratingsAlignVariantsStyle[ratingsAlign],
+          { "items-center": align === "center" }
+        )}
+      >
+        <p className="text-base text-black empty:hidden">{rate}</p>
```
pauzim, line 95: Let's `rate.toFixed(1)` (if rate is defined) according to num format presented in designs

packages/design-system/src/components/Rating/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <StarSvg
+            id={index}
+            active={classNames(showHalfRate(index) && "#FACC15")}
+          />
+        );
+      case "heart":
+        return (
+          <HeartSvg
+            id={index}
+            active={classNames(showHalfRate(index) && "#FACC15")}
```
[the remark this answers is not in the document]
pauzim: EDIT: Since we are using hardcoded color in here I think we could enable dev to pass custom one or be more strict and enable few color variants wdyt?

packages/design-system/src/stories/atoms/Rating.stories.tsx
mzyznowska: Yes, it's required in this case. I was considering this, and despite this disadvantage, it seems to me a better solution than passing a string.

pauzim (approved):

PR #46, opened 2023-03-27
ELPDES-180 - side panel

mzyznowska:


packages/design-system/src/components/SidePanel/index.tsx
```diff
@@ -0,0 +1,62 @@
+import { FC, PropsWithChildren } from "react";
+import IconButton from "@components/IconButton";
+
+import { TComponentIcon } from "@helpers/types";
+import CloseSvg from "@svgs/CloseSvg";
+import classNames from "classnames";
+
+export interface SidePanelProps extends PropsWithChildren {
```
pauzim, line 8: You can export this interface to a separate file to keep consistency with other components

pauzim (approved):
LGTM! I've noticed, that in designs there is an element representing the footer as component's part but I think your approach not defining it as strict flagged element is more flexible :)

PR #66, opened 2023-03-31
Ds 339 carousel

kielbasa-elp:


packages/design-system/src/svgs/ChevronLeftSvg.tsx
```diff
@@ -0,0 +1,23 @@
+import { ClassName } from "@helpers/types";
```
pauzim, line 1: For both Chevron Left and Chevron Right we can use existing `ChevronDownSvg` and rotate it 😃

packages/design-system/tailwind.config.js
```diff
@@ -86,18 +86,23 @@ module.exports = {
         xxs: "10px",
       },
       width: {
+        smActiveCarouselDot: "24px",
```
pauzim, line 89: I think `w-6` tailwind class is a predefined equivalent for `24px`

pauzim (approved):
LGTM, Works as expected!

PR #68, opened 2023-04-03
DS-279 - review component

mzyznowska:


packages/design-system/src/components/Review/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  list,
+  heading,
+  description,
+  icon,
+  rate,
+  maxRate,
+  precision,
+  ...avatarProps
+}) => {
+  const ratingProps = { icon, rate, maxRate, precision };
```
kielbasa-elp, line 53: There are ratingIcon and ratingSize props in the storybook but they dont work for me

kielbasa-elp (approved):
Overall looks good! Just few question

PR #71, opened 2023-04-04
Ds 283 navbar

kielbasa-elp:


packages/design-system/src/stories/molecules/Navbar/AvatarNavbar.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const Avatar: ComponentStory<React.FC<NavbarProps>> = ({ ...props }) => {
+  return (
+    <div className="mx-auto h-[500px] w-full max-w-5xl bg-neutral-50">
+      <Navbar
+        {...props}
+        leftContent={<img src="/logo.png" alt="logo" />}
+        onMenuClick={() => alert("CLICK")}
+      >
+        <div className="flex items-center justify-end gap-2">
+          <div className="flex hidden w-full items-center gap-10 pl-14 lg:flex ">
```
pauzim, line 17:
```suggestion
          <div className="hidden w-full items-center gap-10 pl-14 lg:flex ">
```

packages/design-system/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    return leftContent;
+  }, [leftContent]);
+
+  const renderMenuIcon = useCallback(() => {
+    if (!menuIcon) return null;
+    if (typeof menuIcon === "object") return menuIcon;
+
+    return <MenuButton onClick={onMenuClick} className={menuClassName} />;
+  }, [menuIcon, onMenuClick, menuClassName]);
```
pauzim, line 37: What if we pass `menuIcon` as `MenuButton` prop and render instead of `<MenuSvg/>` ? (in this scenario `MenuSvg` will be the default icon). If it doesn't broke the logic then we can get rid of `typeof` checking

pauzim (approved):
LGTM!

PR #72, opened 2023-04-04
DS-325: `Table` component

pauzim:
Footer element isn't separated because I haven't found any nice solution for making it generic and straightforward yet


kielbasa-elp (approved):
LGTM!

PR #81, opened 2023-04-06
Ds 339 carousel changes

kielbasa-elp:


packages/design-system/src/components/Carousel/carousel.utils.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+type TWrapper = HTMLElement | HTMLDivElement | undefined | null;
+
+export const getSnapWidth = (
+  wrapper: TWrapper,
+  childLength: number,
+  factor = 1
+) => {
+  if (!wrapper) return undefined;
```
pauzim, line 19: It's only my preference, but small refactor to consider: when the function has let's say > 2 arguments I prefer passing a `data` object instead of list of arguments. I think it's in some way more descriptive in places where this function is invoked.

pauzim (approved):
Each carousel variant works great!

PR #83, opened 2023-04-07
Close dropdown on option select

pauzim:


kielbasa-elp (approved):
Overall looks good. Just this one thing with custom triggerElement what we talked about

PR #87, opened 2023-04-11
DS-317 - input suffix icon 

mzyznowska:


pauzim (approved):
I think it's more reusable now

PR #88, opened 2023-04-11
Components statuses

mzyznowska:
I changed statuses of components that are in the column "done" in Jira. Please look at these components and check if you agree with these changes.


pauzim (approved):
Correct my pov 😃

PR #94, opened 2023-04-12
button tokens

piotrekzawada:


pauzim (approved):
LGTM - There is one variant which is not tokenized - `basic` - not sure if should be? (This is the additional variant added during developing `DatePicker` component. In Figma it isn't separated as new variant )

PR #96, opened 2023-04-13
DS-354 - input code

mzyznowska:


packages/design-system/src/components/InputCode/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const generateInputRefsArray = () =>
+    Array.from(Array(lengthOfCode).keys(), () => createRef<HTMLInputElement>());
+
+  const [inputRefsArray, setInputRefsArray] = useState(() =>
+    generateInputRefsArray()
+  );
+
+  useEffect(() => {
+    setInputRefsArray(() => generateInputRefsArray());
+  }, [lengthOfCode]);
```
pauzim, line 27:
```suggestion
  useEffect(() => {
    setInputRefsArray(() => generateInputRefsArray());
  }, [lengthOfCode, generateInputRefsArray]);
```

packages/design-system/src/components/InputCode/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  lengthOfCode = 4,
+  disabled,
+  errorMessage,
+}) => {
+  const [values, changeValue] = useState<{ [key: number]: string }[]>([]);
+  const isError = !!errorMessage;
+
+  const generateInputRefsArray = () =>
+    Array.from(Array(lengthOfCode).keys(), () => createRef<HTMLInputElement>());
+
```
pauzim, line 20:
```suggestion
  const generateInputRefsArray = useCallback(
    () =>
      Array.from(Array(lengthOfCode).keys(), () =>
        createRef<HTMLInputElement>()
      ),
    [lengthOfCode]
  );
```

pauzim (commented):
I think we should add handlers to this component (eg like in `Input.stories.tsx`), to get to the value from outside. For now we cannot get typed value.

pauzim:
One thing to think about - is there a way to pass ref to this input(s) ? 🤔 Not sure if it's crucial but let's assume we are using this as form element - at the end we want to retrieve the value as one part

pauzim (approved):
LGTM 💪 I think we should export `useCodeInput` hook

PR #101, opened 2023-04-14
inline message tokens

piotrekzawada:


pauzim (approved):
Looks good, please look at the Inline message in `error` variant - it looks like the colors weren't found in the palette (?)

PR #102, opened 2023-04-14
DS-341: `Image` component

pauzim:


packages/design-system/src/stories/examples/Image/ImageAsBackground.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <>
+      <small className="mb-2 block text-neutral-600">
+        Please note this is only the example how you can use Tailwind CSS
+        classes to style your image
+      </small>
+      <div className={className} />
+
+      <div className="pt-2">
+        Tailwind className used 👉
+        <span className="pl-2 text-neutral-600">"{className}"</span>
```
mzyznowska, line 46: I think we don't need descriptions under images because in Docs in storybook, we can see everything in the code by clicking `show code`:

Screenshot 2023-04-14 at 16 57 36

mzyznowska (approved):
LGTM 😊 Just think about the descriptions. For me, classNames in the code under the component are enough.

PR #110, opened 2023-04-17
DS-309: BackToTop button

pauzim:
Awaiting for info for variant ghost button text color


kielbasa-elp (approved):
In overall looks good but i am thinking about 3 things: 
1. Component contains few elements like IconButton or Button. Shouldn't it be placed in Molecules?
2.  We check scroll position in the document but what in situations when the scrolled wrapper would be a div or something else inside DOM? Maybe it's good idea to allow pass optional param where we could pass HTMLElement to receive scrollTop from that?
3. Let's add className to wrapper in case we want to change position for example

kielbasa-elp (approved):
LGTM!

PR #111, opened 2023-04-18
DS-377: Select dropdown improvements (input styles)

zuzannanap:


packages/design-system/src/components/Dropdown/Dropdown.interface.ts
```diff
@@ -25,3 +27,8 @@ export type TDropdownVariant =
   | "toggle"
   | "checkbox-left"
   | "checkbox-right";
+
+export interface IDropDownErrorProps {
```
kielbasa-elp, line 32: IDropdownErrorProps

packages/design-system/src/components/Dropdown/Dropdown.interface.ts
```diff
@@ -25,3 +27,8 @@ export type TDropdownVariant =
...
+  message: string;
+  type: "label" | "supportingText" | "errorMessage";
```
kielbasa-elp, line 34: Maybe its good idea to create separate type for that? eg. ` type DropdownErrorType= "label"  | ...`

kielbasa-elp (approved):
Just few changes

PR #114, opened 2023-04-18
DS-391-dropdown-disabling

zuzannanap:


packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx
```diff
@@ -144,11 +141,13 @@ export const SelectDropdownComponents = {
...
-              Control: (props) =>
-                SelectDropdownComponents.Control({
-                  ...props,
-                  isError,
-                }),
+              Control: (props) => (
+                <SelectDropdownComponents.Control
+                  {...props}
+                  isError={isError}
+                  isDisabled
```
kielbasa-elp, line 148: i guess it should be `isDisabled={isDisabled}` :D

kielbasa-elp (approved):

PR #117, opened 2023-04-18
DS-387: Input component fixes

zuzannanap:


packages/design-system/src/components/Input/index.tsx
```diff
@@ -273,7 +273,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
     const isError = !!errorMessage;
 
     return (
-      <>
+      <div>
```
kielbasa-elp, line 276: i am not sure about placing input in additional div. Maybe we can place error message inside label ? We show error instead of supporting text so maybe its good idea to render it there? @pauzim what do you think ?

kielbasa-elp (approved):

packages/design-system/src/components/Input/index.tsx
pauzim: @zuzannanap @kielbasa-elp agree, let's leave `<>` as it was and move error to the `BaseInput`

PR #118, opened 2023-04-18
DS-392:add onInputChange and clearing select

zuzannanap:


packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx
```diff
@@ -122,6 +123,9 @@ export const SelectDropdownComponents = {
             isClearable={false}
             placeholder={placeholderText}
             onChange={(items) => onSelect(items)}
+            onInputChange={(inputValue) => {
+              onInputChange(inputValue);
+            }}
             isDisabled={isDisabled}
```
pauzim, line 129: What this `onInputChange` method does vs `onChange`? 🤔 In storybook I can't see `console.log` called from this method

packages/design-system/src/stories/atoms/Dropdown/SelectDropdown.stories.tsx
```diff
@@ -68,6 +68,8 @@ export default {
     onSelect: (
       value: SingleValue<IDropdownOption> | MultiValue<IDropdownOption>
     ) => console.log("value selected", value),
+    onInpuChange: (inputValue: string) =>
+      console.log("current input value: " + inputValue),
   },
```
pauzim, line 73:
```suggestion
    onInputChange: (inputValue: string) =>
      console.log("current input value: " + inputValue),
  },
```

packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx
```diff
@@ -140,17 +162,31 @@ export const SelectDropdownComponents = {
...
                 <SelectDropdownComponents.Control
                   {...props}
                   isError={isError}
                   isDisabled={isDisabled}
                 />
               ),
+              ClearIndicator: (props) => (
+                <SelectDropdownComponents.ClearIndicator {...props} />
+              ),
             }}
```
pauzim, line 181: Please check if it is necessary to implement `ClearIndicator` that way(inline). It is not recommended by docs https://react-select.com/components (it causes many unnecessary re-renders) and if there aren't custom props we can implement this similar to eg `MultiValueRemove`

packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx
```diff
@@ -140,17 +162,31 @@ export const SelectDropdownComponents = {
...
             styles={customDropdownStyles}
           />
+          <div
+            className={classNames(
+              !supportingText && "empty:min-h-6 pt-1.5",
+              isError && "pt-1.5"
+            )}
+          ></div>
```
pauzim, line 189: @kielbasa-elp what is the current approach - do we leave empty space to prevent input jumping text because I don't remember :( ? cc: @zuzannanap

packages/design-system/src/stories/atoms/Dropdown/SelectDropdown.stories.tsx
pauzim: @zuzannanap there is still a typo in ` onInpuChange` if you leave this as it is this function won't be ever called :P

packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx
zuzannanap: i'll check, but we don't want to use it only with multivalue

zuzannanap: image

pauzim: I meant: can we refactor this to:
```ClearIndicator: SelectDropdownComponents.ClearIndicator```

Comment (zuzannanap) [reply packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx:181]: aaahh ok

Comment (zuzannanap) [reply packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx:181]: ok it works :)

File: packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx (hunk 4 above)
Comment (kielbasa-elp) [reply packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx]: @pauzim no, we don't

File: packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx [hunk 5]
```diff
@@ -107,6 +115,17 @@ export const SelectDropdownComponents = {
...
 
+      const [isSelected, setIsSelected] = useState<boolean>(false);
+
+      const renderDropdownIndicator = useMemo(() => {
+        return (props: IDropdownIndicatorProps) => (
+          <SelectDropdownComponents.Indicator
+            {...props}
+            isHidden={isClearable && isSelected}
+          />
+        );
```
kielbasa-elp, line 126: Wtihout this additional Arrow Func. Just `useMemo((props)=><SelectDropdown...` should work

packages/design-system/src/components/Dropdown/SelectDropdown/index.tsx
```diff
@@ -244,31 +271,41 @@ export const SelectDropdownComponents = {
...
+  }: IDropdownIndicatorProps<IDropdownOption>) =>
+    !isClearable ? (
       <DropdownIndicator {...props}>
         <ChevronDownSvg className="aspect-square w-4 text-black" />
       </DropdownIndicator>
-    );
-  },
+    ) : (
+      <></>
+    ),
```
kielbasa-elp, line 284: maybe return just `null` here if it's possible

packages/design-system/src/components/Input/index.tsx
```diff
@@ -306,8 +306,9 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
             id={ariaDescribedBy}
           />
         )}
-        <div className={!supportingText ? "empty:min-h-6 pt-1.5" : ""} />
-
+        {!supportingText && !isError && (
+          <div className="min-h-6 col-start-1 -mt-1.5 pt-1.5" />
+        )}
```
kielbasa-elp, line 311: Just question, what is purpose of this div ?

kielbasa-elp (approved):

PR #121, opened 2023-04-19
Fix: DS-388 fix workspace item

pauzim:


packages/design-system/src/components/Items/Items.interface.tsx
```diff
@@ -23,7 +23,8 @@ export interface MenuItemProps {
 }
 
 export interface WorkspaceItemProps
-  extends Pick<AvatarProps, "shape" | "src" | "contentType" | "alt"> {
-  size?: BaseComponentSizes;
+  extends Omit<AvatarProps, "label" | "caption"> {
+  itemSize?: BaseComponentSizes;
   variant?: ItemVariants;
+  suffixIcon?: ReactNode;
```
mzyznowska, line 29: Shouldn't `suffixIcon` have `TComponentIcon` instead of `ReactNode` type?

mzyznowska (approved):

PR #122, opened 2023-04-19
DS-369 - input quantity

mzyznowska:


packages/design-system/src/components/InputQuantity/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <div>
+        <Input.BaseInput
+          className="apperance-none z-10 rounded-none border-x [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-outer-spin-button]:appearance-none"
+          prefixVariant="dropdown"
+          suffixVariant="dropdown"
+          ref={combinedRef}
+          isError={isError}
+          disabled={disabled}
+          type="number"
+          onChange={() => {}}
```
pauzim, line 66: Should this function be empty? 🤔

packages/design-system/src/stories/atoms/InputQuantity.stories.tsx
```diff
@@ -0,0 +1,42 @@
+import { Controls, Primary, Title } from "@storybook/addon-docs";
+import InputQuantityComponent from "@components/InputQuantity";
+import type { InputQuantityProps } from "@components/InputQuantity/InputQuantity.interface";
+import { useRef } from "react";
+
+export const InputQuantity = ({ ...props }: InputQuantityProps) => {
+  const inputRef = useRef(null);
+
+  return <InputQuantityComponent {...props} ref={inputRef} />;
+};
```
pauzim, line 10: We need a way to get input value from outside the component

packages/design-system/src/components/InputQuantity/index.tsx
```diff
@@ -53,6 +53,7 @@ const InputQuantity = forwardRef<HTMLInputElement, InputQuantityProps>(
         const count = Number(defaultRef.current.value);
         const newValue = variant === "increase" ? count + 1 : count - 1;
         defaultRef.current.value = newValue.toString();
+        getValue && getValue(newValue.toString());
```
pauzim, line 56: Suggestion: Can be simplified to `getValue?.(newValue.toString())`

packages/design-system/src/components/InputQuantity/index.tsx
```diff
@@ -66,6 +67,7 @@ const InputQuantity = forwardRef<HTMLInputElement, InputQuantityProps>(
           isError={isError}
           disabled={disabled}
           type="number"
+          onChange={(event) => getValue && getValue(event.target.value)}
```
pauzim, line 70: Suggestion: Also can be simplified to `onChange={(event) => getValue?.(event.target.value)}`

packages/design-system/src/components/InputQuantity/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { FC, forwardRef, useRef } from "react";
+import { FC, forwardRef, useEffect, useRef, useState } from "react";
```
pauzim, line 1: I think `useEffect` and `useState` are unused vars

pauzim (approved):
LGTM! 🎉 One thing I consider is if we want the value as string or as a number? 🤔

PR #124, opened 2023-04-20
Select dropdown improvements

zuzannanap:


packages/design-system/src/components/Dropdown/SelectDropdown/SelectDropdown.interface.ts
```diff
@@ -13,10 +13,11 @@ export interface ISelectDropdownProps {
   isDisabled?: boolean;
   instanceId?: string;
   isClearable?: boolean;
+  isLoading?: boolean;
+  noOptionsMessage?: string;
   onSelect: (
     items: SingleValue<IDropdownOption> | MultiValue<IDropdownOption>
   ) => void;
-  onInputChange: (inputValue: string) => void;
```
mzyznowska, line 19: I think you shouldn't remove `onInputChange` from interface

packages/design-system/src/stories/atoms/Dropdown/SelectDropdown.stories.tsx
```diff
@@ -58,6 +58,12 @@ export default {
     errorMessage: {
       description: "string",
     },
+    inputValue: {
+      description: "string",
+    },
```
mzyznowska, line 63: I'm not sure if `inputValue` is necessary. It doesn't exist in `ISelectDropdownProps`. You can add it to interface and use here` inputValue: {control: false }` only to show this prop or remove it if we don't need it.

mzyznowska (approved):

PR #126, opened 2023-04-21
checkbox tokens

piotrekzawada:


pauzim (approved):
Looks good - one thing that differs from previous version is `indeterminate` state - in storybook when I set `indeterminate` to true the `-` icon inside checkbox is missing

PR #129, opened 2023-04-23
[POC] DS-305: Charts - Bar Chart with `charts.css` lib

pauzim:
POC if we can use `charts.css` lib for building charts designed in DS.
**Problems not solved:**
- Lib promotes `css` styles, not much space to implement tailwind classes;
- You have to use css variables to set values for the charts;
- Lack of `PieChart` - still under development for at least 2 years;
- There are no possibility to set labels to both of axes (x,y)https://github.com/ChartsCSS/charts.css/issues/19;
- Scale have to be set manually by adding css class


pauzim:
Closed because we decided not to use `charts.css` because of library limitation in charts styling (not possible to add labels to both of axes)

pauzim:
Closed because we decided not to use `charts.css` library because of limitations in charts styling (for example there wasn't any possibility to display labels for both of axes).

PR #133, opened 2023-04-24
DS-368: Multiselect

zuzannanap:
display "+1", "+2" etc. when some items don't fit in the value container


packages/design-system/src/components/Dropdown/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+      el.style.opacity = "0";
+      el.style.pointerEvents = "none";
+      if (!(nextTagsWidth + 120 > containerWidth)) {
+        nextHiddenElements--;
+        el.removeAttribute("style");
+      }
+
+      return { tagsWidth: nextTagsWidth, hiddenElements: nextHiddenElements };
+    },
+    { tagsWidth: 0, hiddenElements: elements.length }
```
kielbasa-elp, line 19: I think we should change "tags.." to something else

packages/design-system/src/utils/useWindowSize.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  const handleSize = () => {
+    setWindowSize({
+      width: window.innerWidth,
+      height: window.innerHeight,
+    });
+  };
+
+  useEffect(() => {
```
kielbasa-elp, line 21: we should also use useIsomorphicLayoutEffect here instead of useEffect

PR #138, opened 2023-04-26
DS-394: Dropdown Select refactor

pauzim:


packages/design-system/src/components/Dropdown/SelectDropdown/SelectDropdown.interface.ts
```diff
@@ -21,6 +32,55 @@ export interface ISelectDropdownProps {
...
+  getOptionLabel?: (option: IDropdownOption) => string;
+  getOptionValue?: (option: IDropdownOption) => string;
+  formatOptionLabel?: (
+    data: IDropdownOption,
+    formatOptionLabelMeta: FormatOptionLabelMeta<IDropdownOption>
+  ) => ReactNode;
+  loadingMessage?: (obj: { inputValue: string }) => ReactNode;
+  onKeyDown?: React.KeyboardEventHandler<HTMLDivElement>;
+  onMenuScrollToTop?: (event: WheelEvent | TouchEvent) => void;
+  onMenuScrollToBottom?: (event: WheelEvent | TouchEvent) => void;
```
zuzannanap, line 81: some of those props might not work as the should (e.g. blurInputOnSelect doesn't work properlu) - i think it would be google to test them all before adding to the component

and another thing is: shouldn't these props be added to storybook as well?

pauzim: My goal was to enable as much `react-select` lib props as possible to enable covering different project business cases scenarios to prevent situation you had when using component in Flounder.  I don't think we have to test them one by one because it's 1:1 copied from lib's interfaces. If some of them doesn't - it's the issue you can address directly to the lib authors.

Same if it comes to storybook - we display controls which cover variants from designed component, mostly from the visual pov.

mzyznowska (approved):

PR #141, opened 2023-04-27
DS-370: Number Input

pauzim:


packages/design-system/src/components/InputNumber/InputNumber.interface.ts
```diff
@@ -0,0 +1,36 @@
...
+    | "type"
+    | "onChange"
+    | "suffixIcon"
+    | "onIconClick"
+    | "isError"
+  > {
+  minValue?: number;
+  maxValue?: number;
+  step?: number;
+  enableNegatives?: boolean;
```
piotrekzawada, line 19: i really don't like the idea of passing boolean as prop - it indicates that there is an if somewhere down the road. I think we can relay on minValue especially it's optional. So let's assume that all input without minValue can be negative

packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+          actionType === EActionType.INCREMENT ? count + step : count - step;
+        const newValue = defineNewValue(calculatedValue);
+        defaultRef.current.value = newValue;
+        onChange?.(newValue);
+      }
+    };
+
+    return (
+      <>
+        {label && <Input.Text text={label} type="label" />}
```
piotrekzawada, line 97: does the `Input.Text` supports `empty:hidden`? If so we don't need this condition

packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+              onClick={() => handleClick(EActionType.DECREMENT)}
+              disabled={disabled}
+              className="rounded-b-[4px]"
+            />
+          </div>
+        </Input.BaseInput>
+
+        {supportingText && !isError && (
+          <Input.Text text={supportingText} type="supportingText" />
+        )}
```
piotrekzawada, line 129: same as above

packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <div className="min-h-6">
+          {isError && <Input.Text text={errorMessage} type="errorMessage" />}
```
piotrekzawada, line 131: same as above

packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    <button
+      onClick={onClick}
+      className={classNames(
+        "flex aspect-square w-3 items-center justify-center bg-neutral-100 hover:bg-neutral-300 hover:text-white",
+        "disabled:cursor-not-allowed disabled:hover:bg-neutral-100 disabled:hover:text-neutral-700",
+        className
+      )}
+      disabled={disabled}
+    >
+      <ChevronDownSvg
```
zuzannanap, line 28: i think chevron icons are a bit to small (smaller than on figma)

packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+      supportingText,
+      label,
+      disabled,
+      ...props
+    },
+    ref
+  ) => {
+    const defaultRef = useRef<HTMLInputElement>(null);
+    const combinedRef = useCombinedRefs(ref, defaultRef);
+    const isError = !!errorMessage;
```
zuzannanap, line 56: box-shadow doesn't change for red when there is an error now

packages/design-system/src/stories/atoms/InputNumber.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    errorMessage: errorMessageControl,
+  },
+  args: {
+    label: "",
+    supportingText: "",
+    defaultValue: 0,
+    disabled: false,
+    errorMessage: "",
+    step: 1,
+    minValue: 0,
```
zuzannanap, line 37: you can't change input value on storybook when you change min or max value (but i guess it's just o storybook problem)

zuzannanap (commented):
i

packages/design-system/src/components/InputNumber/index.tsx
pauzim: Ah good one! it's the bug in main input component, I'll try to fix it 👀

packages/design-system/src/stories/atoms/InputNumber.stories.tsx
pauzim: I've removed `value` control since it's a controlled component 👌

piotrekzawada (approved):

packages/design-system/src/components/InputNumber/index.tsx
pauzim: fixed in https://github.com/elpassion/flounder-2/pull/145

PR #142, opened 2023-04-28
DS-353: Refactor Notifications

pauzim:


kielbasa-elp (approved):
It's look good!

packages/design-system/src/components/Notification/index.tsx
```diff
@@ -1,141 +1,41 @@
...
-          "border-blue-500 bg-blue-500",
-          "hover:border-blue-700 hover:bg-blue-700",
-          "active:border-blue-800 active:bg-blue-800"
-        )}
-      />
-    </footer>
-  </NotificationComponents.Container>
+export const Notification = ({ children }: PropsWithChildren) => (
+  <>{children}</>
 );
```
mzyznowska, line 36: I'm not sure about this, maybe it could be a container with default styles e.g background, shadow... and with className to add custom styles?

zuzannanap (approved):

packages/design-system/src/components/Notification/index.tsx
pauzim: @mzyznowska  To be honest, I've added this `Notification` only because of imports requirements - to be able to export Notification components and reuse them with dot notation 😄 I thought about exporting container with default styles  as separate notification element as it was in previous implementation - in this scenario developer could choose whether to use default container from DS or custom one wdyt?

packages/design-system/src/components/Notification/index.tsx
```diff
@@ -1,141 +1,41 @@
...
+export const NotificationItem: FC<NotificationItemProps> = ({
+  children,
+  isUnread,
+  className,
+}) => (
   <div
     className={classNames(
-      "shadow-elevatio-sm w-full divide-y rounded-lg bg-white",
+      "flex w-full items-start justify-between gap-4 py-3 px-4 text-sm",
       className
```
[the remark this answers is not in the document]
pauzim: I've assumed hover effect can be added to the wrapper element -> in storybook example it's a `button` element. If your business logic requires adding hover effect directly to this wrapper, you can add it with `className`. I think it would be more simple than overriding these styles if necessary.

packages/design-system/src/components/Notification/index.tsx
mzyznowska: I think having a container with default styles would be great, especially since it could be modified with tokens.

pauzim: @mzyznowska done 😃

mzyznowska (approved):

PR #143, opened 2023-04-28
DS-308 - video player

mzyznowska:


zuzannanap (approved):
looks good to me!

i think in the future it would be nice to have those icons in our font too

packages/design-system/src/components/VideoPlayer/VideoPlayer.interface.ts
```diff
@@ -0,0 +1,26 @@
+import { TComponentIcon } from "@helpers/types";
+import { PropsWithChildren } from "react";
+
+type TControlsItem = { icon: TComponentIcon; onClick(): void };
```
kielbasa-elp, line 4: I think that the onClick should returns Event in the param

kielbasa-elp (approved):
In overall its look good but one thing to think about. We do not have ability to pass custom controls, make "Next" button disable or sth like that. I think it would be great to make it possible in the future.

PR #147, opened 2023-04-29
DS-397: WYSIWYG

zuzannanap:


examples/with-design-system/libs/ui/src/components/Wysiwyg/WysiwygEditor.stories.tsx
```diff
@@ -1,35 +1,35 @@
...
+  title: "Organisms/WysiwygEditor",
   component: WYSIWYGEditor,
 } as ComponentMeta<typeof WYSIWYGEditor>;
 
-const Template: ComponentStory<typeof WYSIWYGEditor> = (args) => (
-  <WYSIWYGEditor {...args} />
-);
+const Template: ComponentStory<typeof WYSIWYGEditor> = args => {
+  return <WYSIWYGEditor value={...args} />;
+};
```
pauzim, line 13: 1. Shouldn't we follow the same story pattern as we have in all other components? 🤔

pauzim (commented):
There's sth with the `Toolbar`s rwd -it leaks outside the container:
Screenshot 2023-05-08 at 09 09 04

Also one thing to consider (we can add separate task for it if needed) -> when I type some phrase and remove all eg with backspace key, editor's state isn't empty, it has empty HTML `<p></br></p>` Not sure what's the logic under this in the `quill` lib but it can be possibly an issue when you have to eg check inside form if the field is empty or not to validate.

mzyznowska:
> Also one thing to consider (we can add separate task for it if needed) -> when I type some phrase and remove all eg with backspace key, editor's state isn't empty, it has empty HTML `<p></br></p>` Not sure what's the logic under this in the `quill` lib but it can be possibly an issue when you have to eg check inside form if the field is empty or not to validate.

I looked for a solution, but I couldn't find it. But it's only raw text with HTML. If it is used in a project, writing logic to recognise tags will be necessary. For now, it's the whole value of the editor.

pauzim (approved):

PR #148, opened 2023-05-02
DS-412 - fix inputs

mzyznowska:


packages/design-system/src/components/Input/index.tsx
```diff
@@ -225,6 +225,9 @@ export const Text: React.FC<TextProps> = ({ text, type }) => {
       "col-start-1 col-end-3 text-neutral-400 peer-disabled:text-neutral-200",
     errorMessage: "col-start-1 col-end-3 m-0 p-0 text-error-500",
   };
+
+  if(!text) return null
+
```
pauzim, line 230: I'm not sure if we need this since we return div with `empty:hidden` 🤔

pauzim (approved):

PR #149, opened 2023-05-04
[BUG] DS-417: Input Number & Input Quantity -> fix minValue increment

pauzim:
Handle scenario when `minValue` is set to > 0. In previous implementation increment was blocked.


packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -86,14 +86,14 @@ const InputNumber = forwardRef<HTMLInputElement, InputNumberProps>(
         const calculatedValue =
           actionType === EActionType.INCREMENT ? count + step : count - step;
 
-        if (
-          (minValue && calculatedValue < minValue) ||
-          (maxValue && calculatedValue > maxValue)
-        ) {
+        if (maxValue && calculatedValue > maxValue) {
```
mzyznowska, line 89: It doesn't work when maxValue (or minValue) is 0. In InputQuantity, I check types of values, or maybe you have a better solution?

mzyznowska (approved):

PR #150, opened 2023-05-04
DS-413: Remove circular dependencies

pauzim:


mzyznowska (approved):
LGTM 👌 Only one thing to think ;) There are separate folders for subcomponents for some components, but for the others, you left them in the same file (e.g. Indicator). I think it's because of size of file but maybe we should have one pattern?

pauzim:
@mzyznowska that's a good point - yeah I've leaved some components in a single file because of the sizes ;) But I'm not fully satisfied because of inconsistency you have mentioned. I can move all of them regardless file sizes. WDYT?

PR #157, opened 2023-05-08
Ds 426: Icons classes

zuzannanap:
alternatively, we can add icon names directly in "className", but I think that this way is clearer - I'd like to know your opinion!


packages/design-system/src/components/Icon/index.tsx
```diff
@@ -11,26 +11,42 @@ export const Icon: React.FC<IconProps> = ({
...
 
+  const styleSheet = document.styleSheets[0];
+  Object.keys(allIcons).forEach((icon) => {
+    const selector = `.${icon}:before { content: '${allIcons?.[icon]?.entity}'; display: none; }`;
+    styleSheet.insertRule(selector, styleSheet.cssRules.length);
+  });
+
+  const spanRef = useRef<HTMLSpanElement | null>(null);
+  useEffect(() => {
+    const spanEl = spanRef.current;
```
mzyznowska, line 29:
```suggestion
    const spanEl = spanRef?.current;
```

PR #163, opened 2023-05-09
DS-396  - tooltip tokens

mzyznowska:


pauzim (approved):
Overall works good! Only one difference I've noticed is the behavior of `className` prop. I tried with some bg class. Before tokens, the color have changed, now it's not.

kielbasa-elp (approved):

PR #164, opened 2023-05-09
DS-421: WYSIWYG disabled state

zuzannanap:
it seems to be a storybook issue, but when switching from disabled false to true (and vice versa), styles break... if you change the default to true in the code, everything will be displayed properly


packages/design-system/src/components/WYSIWYGEditor/index.tsx
```diff
@@ -76,43 +79,87 @@ const WYSIWYGEditor = forwardRef<typeof ReactQuill, IWYSIWYGEditorProps>(
...
+            <Suspense fallback="">
+              <ReactQuill
+                modules={modules}
+                value={value}
+                onChange={setValue}
+                placeholder={inputPlaceholder}
+                theme="snow"
+                formats={formats}
+                readOnly={disabled}
+                onFocus={!disabled ? handleInputFocus : () => {}}
```
pauzim, line 146:
```suggestion
                formats={formats}
                readOnly={disabled}
                onFocus={!disabled ? handleInputFocus : undefined}
```

packages/design-system/src/components/WYSIWYGEditor/index.tsx
```diff
@@ -76,43 +79,87 @@ const WYSIWYGEditor = forwardRef<typeof ReactQuill, IWYSIWYGEditorProps>(
...
         <div
-          ref={containerRef}
           className={classNames(
             "wysiwyg-container border-1 flex w-full flex-col rounded-lg border focus:shadow-md",
-            isError ? "border-error-500" : borderVariantsStyle[borderVariant],
+            isError
+              ? "border-error-500"
+              : disabled
+              ? "border-neutral-200"
+              : borderVariantsStyle[borderVariant],
```
pauzim, line 129: Could we extract this to some small function rather than nesting conditions?

pauzim (approved):

kielbasa-elp (approved):

PR #172, opened 2023-05-10
fix/DS-428 - datepicker validation

mzyznowska:


pauzim (commented):
I have some issue when the date format is set to default `mm/dd/yyyy` I can't type `3` in day position (but this appears only in `startDate` in `endDate` I can type this) 🤔

On the other hand I'm not sure if it's worth it to write these complicated validations for each date format- it looks like it's very hard to avoid some edge cases. As an alternative we could validate typed data and return error if it's not valid. I know it's not perfect but maybe good enough for now ;)

zuzannanap:
image
enddate is not validated and btw it allows to type only 7 numbers :<

zuzannanap:
not related to your task: enddate allows to type only 7 numbers :<

zuzannanap:
> I have some issue when the date format is set to default `mm/dd/yyyy` I can't type `3` in day position (but this appears only in `startDate` in `endDate` I can type this) 🤔
> 
> On the other hand I'm not sure if it's worth it to write these complicated validations for each date format- it looks like it's very hard to avoid some edge cases. As an alternative we could validate typed data and return error if it's not valid. I know it's not perfect but maybe good enough for now ;)

didn't you try it for february?
it works good to me

pauzim:
> > I have some issue when the date format is set to default `mm/dd/yyyy` I can't type `3` in day position (but this appears only in `startDate` in `endDate` I can type this) 🤔
> > On the other hand I'm not sure if it's worth it to write these complicated validations for each date format- it looks like it's very hard to avoid some edge cases. As an alternative we could validate typed data and return error if it's not valid. I know it's not perfect but maybe good enough for now ;)
> 
> didn't you try it for february? it works good to me

I think actually February can't have more than 29 days so `3` should be falsy? 🙃

mzyznowska:
> > > I have some issue when the date format is set to default `mm/dd/yyyy` I can't type `3` in day position (but this appears only in `startDate` in `endDate` I can type this) 🤔
> > > On the other hand I'm not sure if it's worth it to write these complicated validations for each date format- it looks like it's very hard to avoid some edge cases. As an alternative we could validate typed data and return error if it's not valid. I know it's not perfect but maybe good enough for now ;)
> > 
> > 
> > didn't you try it for february? it works good to me
> 
> I think actually February can't have more than 29 days so `3` should be falsy? 🙃

@pauzim @zuzannanap 
So, now if you have `02/`, you can't write more than 29, and there are also rules for long and short months. I'm not sure if it's the best user experience when you can't write. Maybe we should let the user write all numbers and show the red border when something is wrong?

zuzannanap:
> > > I have some issue when the date format is set to default `mm/dd/yyyy` I can't type `3` in day position (but this appears only in `startDate` in `endDate` I can type this) 🤔
> > > On the other hand I'm not sure if it's worth it to write these complicated validations for each date format- it looks like it's very hard to avoid some edge cases. As an alternative we could validate typed data and return error if it's not valid. I know it's not perfect but maybe good enough for now ;)
> > 
> > 
> > didn't you try it for february? it works good to me
> 
> I think actually February can't have more than 29 days so `3` should be falsy? 🙃

so that't why i'm asking - i thought maybe you checked it for february unwittingly ;)


<br class="Apple-interchange-newline">

kielbasa-elp (commented):
Just suggestion because i don't know this component and the issue well. 
Maybe it would be easier and simpler to validate date not on input change but onBlur?
The validation on input change is tricky and i am not sure if its good for UX

packages/design-system/src/components/DatePicker/DatePicker.tsx
```diff
@@ -33,6 +33,7 @@ const DatePicker: FC<DatePickerProps> = ({
 }) => {
   const [isMonthPickerActive, setIsMonthPickerActive] = useState<boolean>();
   const [isYearPickerActive, setIsYearPickerActive] = useState<boolean>();
+  const [isError, setIsError] = useState<boolean>();
```
kielbasa-elp, line 36: Set false as default of useState

kielbasa-elp (approved):
LGTM! GJ
I am just thinking because i see that we are using date-fns in few places in DesignSystem.
Maybe we can use parse and isValid functions (from date-fns package) to validate date instead of writing validation ourself?

PR #174, opened 2023-05-11
[WiP] DS-419: Async select

pauzim:


packages/design-system/src/components/Dropdown/AsyncSelect/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+      isError,
+      isSelected,
+      variant,
+    });
+
+    return (
+      <div>
+        {hasLabel && (
+          <SelectDropdownComponents.Text message={label} type="label" />
+        )}
```
mzyznowska, line 60: Can we add `empty:hidden` to <SelectDropdownComponents.Text/> ? And the same with supporting text and error like in Inputs?

packages/design-system/src/stories/atoms/Dropdown/AsyncSelect.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+const option = [{ label: "pomidor", id: "1245", value: "pomidor" }];
+const defaultOption = { option };
+
+export const AsyncSelect = ({ options, ...props }: IAsyncSelectProps) => {
+  const [selectedItems, setSelectedItems] = useState<
+    SingleValue<IDropdownOption> | MultiValue<IDropdownOption>
+  >();
+
+  console.log("Select Dropdown - selectedItems", selectedItems);
```
mzyznowska, line 20:
```suggestion
```

mzyznowska (commented):
I don't see any options, even if I choose something from the dropdown with options in args.

packages/design-system/src/components/Dropdown/AsyncSelect/index.tsx
pauzim: Yes, good point! 🙏 (Also fixed in `SelectDropdown` component

pauzim:
> I don't see any options, even if I choose something from the dropdown with options in args.

Have you tried to type some option? because this select by default behaves more as autocomplete. 

I fixed `defaultOptions` control - if you select `true` and refresh storybook window (didn't find any workaround for this),  the options will be loaded and visible initially when component will be rendered (you will see some `...` as the proof that loading option promise is pending 😃

mzyznowska (approved):

PR #187, opened 2023-05-16
DS-453: fix toggle in storybook

zuzannanap:


packages/design-system/src/stories/atoms/Toggle.stories.tsx
```diff
@@ -12,9 +12,25 @@ const docs: string = `# Usage <br/>
...
-  <ToggleComponent {...props} />
-);
+export const Toggle = ({ ...props }: ToggleProps) => {
+  const [checked, setChecked] = useState<boolean>(false);
+  const [message, setMessage] = useState<string>("Not checked");
+  return (
+    <>
+      <ToggleComponent
+        {...props}
+        checked={checked}
```
mzyznowska, line 22: We don't have to pass `checked` here :)

mzyznowska (approved):

packages/design-system/src/stories/atoms/Toggle.stories.tsx
zuzannanap: yeah, you're right!

PR #188, opened 2023-05-16
Ds 434 hero banner

zuzannanap:


packages/design-system/src/stories/molecules/HeroBanner/HeroBanner.stories.tsx
```diff
@@ -0,0 +1,70 @@
+import HeroBannerComponent from "@components/HeroBanner";
+import { HeroBannerProps } from "@components/HeroBanner/HeroBanner.interface";
+import { loopOptions } from "stories/utils";
+
+export const HeroBanner = ({ ...props }: HeroBannerProps) => (
+  <HeroBannerComponent {...props} />
+);
+
+export default {
+  title: "🟢 Molecules/HeroBanner",
```
mzyznowska, line 10:
```suggestion
  title: "🟠 Molecules/HeroBanner",
```

mzyznowska:
And loop doesn't work for me. I found that it could be something with the slides' width.

packages/design-system/src/components/Carousel/Arrows/CarouselArrows.tsx
```diff
@@ -61,7 +61,7 @@ export const NextButton: FC<CarouselButtonProps> = (props) => {
       disabled={!enabled}
       className={classNames(
         buttonStyles,
-        "-right-6",
+        "right-[5%]",
```
kielbasa-elp, line 64: Do not change button styles. Instead of pass arrow buttons directly to the component with the new styles.

kielbasa-elp (changes requested):
We have to create this component in different way because for now too much content is hardcoded. 

Maybe instead of preparing strict Slides, lets create just HeroBanner.Slide component which will be just wrapper for content with correct aspect ratio. Everything else would be define by users directly during creating slides.

mzyznowska (approved):

PR #191, opened 2023-05-16
DS-459: add anchor tokens

zuzannanap:


packages/design-system/src/helpers/tokens/anchor.json
```diff
@@ -0,0 +1,76 @@
...
+          "type": "color"
+        },
+        "focus": {
+          "value": "{primary.300}",
+          "type": "color"
+        }
+      }
+    }
+  },
+  "bottomDecoration": {
```
mzyznowska, line 20:
```suggestion
  "decorationBottom": {
```
I think it doesn't work ;) In Anchor component you have `decorationBottom`.

PR #193, opened 2023-05-17
Ds 461 back to bottom button tokens

zuzannanap:


packages/design-system/src/helpers/defaultTheme.js
```diff
@@ -42,6 +43,11 @@ const tooltip = tokens.getComponentColorConfig(
   "tooltip",
   colorPalette
 );
+const backToTopButton = tokens.getComponentColorConfig(
+  backToTopButtonColors,
+  "backToTopButton",
+    colorPalette
```
mzyznowska, line 49: Format this file before merge :)

mzyznowska (approved):

PR #195, opened 2023-05-17
DS-462-breadcrumbs colors tokens

zuzannanap:


packages/design-system/src/helpers/tokens/breadcrumbs.json
```diff
@@ -0,0 +1,77 @@
...
+      "color": {
+        "border": {
+          "default": {
+            "value": "{neutral.100}",
+            "type": "color"
+          }
+        }
+      }
+    },
+    "backgroundVariant": {
```
mzyznowska, line 62: Couldn't we have `background` here instead of `backgroundVariant`?

zuzannanap: ok!

PR #207, opened 2023-05-23
DS-475 - separate components in package

mzyznowska:


kielbasa-elp (changes requested):
Add components to "components" array in src/helpers/generatePaths.js. 

Also I am afraid that we will have to change index.tsx to idnex.ts

PR #209, opened 2023-05-23
Ds 450 iconButton colors tokens

zuzannanap:


packages/design-system/src/components/IconButton/index.tsx
```diff
@@ -15,19 +15,19 @@ export const IconButton: React.FC<IconButtonProps> = ({
 }) => {
   const styleVariants: GenericKeys = {
     primary:
```
mzyznowska, line 17: I changed it to `filled` variant. For now, I don't know what we should do with tokens. In the button I didn't change `primary` tokens because this version is still in figma.

mzyznowska (approved):

PR #217, opened 2023-05-24
DS-492: InputQuantity colors tokens

zuzannanap:


kielbasa-elp (approved):
Tokens works but i think that there is sth wrong with labels, errors etc. Error and supporting text are above component

PR #224, opened 2023-05-26
DS-497 menuItems colors tokens

zuzannanap:


kielbasa-elp (changes requested):
isActive prop doesn't work

kielbasa-elp (approved):

PR #227, opened 2023-05-26
DS-503: pagination colors tokens

zuzannanap:


kielbasa-elp (changes requested):
I am not sure if text-pagination-base works :v buttons have black text color which is incorrect right?

kielbasa-elp (approved):

PR #230, opened 2023-05-29
DS-502: Notifications colors tokens

zuzannanap:


kielbasa-elp (commented):
Works good. One question - you removed the background of EmptyState? Why is that ?

zuzannanap:
> Works good. One question - you removed the background of EmptyState? Why is that ?

cause it looked like this

image
instead of this:
image

PR #231, opened 2023-05-29
DS-501: NavTabs colors tokens

zuzannanap:


packages/design-system/src/components/Icon/index.tsx
```diff
@@ -25,7 +25,7 @@ export const Icon: FC<IconProps> = ({
     return <>{customIcon}</>;
   }
   const iconCode = convertHtmlEntityToUnicode(entity);
-
+  const iconEntityCode = convertHtmlEntityToUnicode(iconEntity);
```
kielbasa-elp, line 28: Maybe just
` const iconCode = convertHtmlEntityToUnicode(entity || iconEntity);`

packages/design-system/src/components/Icon/index.tsx
```diff
@@ -35,7 +35,7 @@ export const Icon: FC<IconProps> = ({
       )}
       {...rest}
     >
-      {iconCode}
+      {iconCode || iconEntityCode}
```
kielbasa-elp, line 38: and then remove OR operator from here

kielbasa-elp (approved):

PR #241, opened 2023-05-30
DS-450 Stepper colors tokens

zuzannanap:


packages/design-system/src/components/Stepper/Stepper.interface.ts
```diff
@@ -33,18 +33,22 @@ export interface TitleSectionProps {
...
   icon?: TComponentIcon;
   completeVariant?: "default" | "check";
+  stepClassName?: string;
 }
 
 export interface StepperProps
   extends Pick<StepProps, "size" | "onClick" | "completeVariant"> {
   steps: StepInterface[];
   activeStep: number;
+  className?: string;
```
kielbasa-elp, line 52: instead of manually defining className prop, extend StepperProps with ClassName
```
export interface StepperProps
  extends Pick<StepProps, "size" | "onClick" | "completeVariant"> , ClassName {...}
```
You can import ClassName from  `"@helpers/types";`

kielbasa-elp (approved):

PR #246, opened 2023-05-30
DS-516: TextArea colors tokens

zuzannanap:


kielbasa-elp (approved):
looks good! 
One thing not related to tokens. When textarea has error the border changes to red which is correct but the Ring is still purple

PR #247, opened 2023-05-30
DS-470: Dropdown colors tokens

zuzannanap:


kielbasa-elp (changes requested):
Take a look at Async and Select Dropdowns. Focus at select - the outline is red but shouldn't

PR #248, opened 2023-05-30
Ds 524: Optional radio props + style fix

zuzannanap:


pauzim (commented):
@zuzannanap I forgot, please pass `...props` to the input and extend
 ```
<input  type="radio" ...
{...props} />

```
Because currently I'm not able to get eg value of the input

kielbasa-elp (approved):

PR #250, opened 2023-06-01
DS-525 - creatable select

mzyznowska:


zuzannanap:
looks good, but it will be nice to have the ability to change "Create" text if it's possible

zuzannanap (approved):

PR #255, opened 2023-06-02
DS-530/part 2 - border radius tokens

mzyznowska:


zuzannanap:
left side border radiuses are wrong :<
image

zuzannanap:
In inutNumber borderRadiuses of arrows should be rounded (they weren't, it's not tokens problem)
figma: 
image

zuzannanap:
radiuses in Legend don't work :<

zuzannanap:
SmallToast borders don't work
image

packages/design-system/src/components/Toast/index.tsx
```diff
@@ -76,7 +76,7 @@ export const ToastComponents = {
   Container: ({ children, className }: ContainerProps) => {
     return (
       <div
-        className={classNames("rounded-lg shadow-xl empty:hidden", className)}
+        className={classNames("rounded-toast-baseRadius shadow-xl empty:hidden", className)}
```
zuzannanap, line 79: i guess it doesn't work cause it should be 'radiusBase' ;)

zuzannanap (changes requested):

mzyznowska:
> left side border radiuses are wrong :< image

It is correct in my storybook 😅

zuzannanap:
> > left side border radiuses are wrong :< image
> 
> It is correct in my storybook 😅

ok, now it's also ok in my storybook...🙃

zuzannanap (approved):

PR #256, opened 2023-06-02
DS-533 - clean storybook

mzyznowska:


packages/design-system/src/helpers/tokensHelpers/colorsTokensHelper.js
```diff
@@ -1,4 +1,4 @@
-const colors = require("../tokens/colors.json");
+const colors = require("../tokens/tokensPalettes/colors.json");
```
zuzannanap, line 1: is "Palletes" a good word for borderRadius?

mzyznowska: What about "tokensConfigs"?

zuzannanap (approved):

PR #262, opened 2023-06-05
fix/DS-540 - creatable new options

mzyznowska:
I decided to add a hook. Developers can use it or write logic and functions themselves.


pauzim (approved):
Haven't tested it but looks good. I think hook is the great idea! 🎉

PR #264, opened 2023-06-06
DS-529 Setup Loki Visual Regression Testing

zuzannanap:


.github/workflows/test-visual-regression.yml
```diff
@@ -0,0 +1,59 @@
...
+          fetch-depth: 0
+        steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.18.1
```
juliaszymura, line 27: Is this exact version required? This will require more maintenance compared to major version only.

packages/design-system/loki.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  configurations: {
+    "chrome.laptop": {
```
juliaszymura, line 3: Wouldn't you like to have a mobile-like reference as well?

.github/workflows/update-visual-regression-baseline-manually.yml
```diff
@@ -0,0 +1,44 @@
+name: 'Update visual regression baseline'
```
juliaszymura, line 1: If you go with the diff pr merge workflow to update baseline images I think this one will not be used as well

juliaszymura (commented):
Great job 🔥 It would be awesome to have this kind of testing documented in readme as well.

packages/design-system/loki.config.js
zuzannanap: we thought it's not needed for now as DS doesn't have mobile version - do ypu think we should test it on mobile version anyway?

juliaszymura: If DS doesn't have components with different layouts on mobile I think this is unnecessary at the moment

PR #265, opened 2023-06-06
Ds 531 avatar status

zuzannanap:


packages/design-system/src/components/Avatar/AvatarComponents.tsx
```diff
@@ -24,18 +27,27 @@ const AvatarComponents = {
...
+          )}
+        >
+          {children}
+        </div>
+        {!!statusSign && (
+          <AvatarComponents.Status
+            position={shape === "square" ? "top" : "bottom"}
+            size={size}
+            children={statusSign}
+          />
```
monikaracka, line 49: It shouldn't be part of Container in AvatarComponents because there are in this file separate components that you can use to create your own avatar. You should move this to index.tsx.

packages/design-system/src/components/Avatar/Avatar.interface.ts
```diff
@@ -7,10 +7,12 @@ import type {
 
 type Sizes = "xxs" | "xs" | BaseComponentSizes | "xl" | "xxl";
 type Shapes = "circle" | "square";
+type StatusPositions = "top" | "bottom";
 
 export interface ContainerProps extends PropsWithChildren, ClassName {
   shape?: Shapes;
   size?: Sizes;
+  statusSign?: ReactNode;
```
monikaracka, line 15: We don't need this prop here.

packages/design-system/src/components/Avatar/AvatarComponents.tsx
```diff
@@ -24,18 +27,27 @@ const AvatarComponents = {
...
+              "rounded-lg": shape === "square",
+            },
+            className
+          )}
+        >
+          {children}
+        </div>
+        {!!statusSign && (
+          <AvatarComponents.Status
+            position={shape === "square" ? "top" : "bottom"}
```
monikaracka, line 46: I know how examples with statuses look in Figma, but maybe we should make it possible to choose the position regardless of the shape? What do you think?

packages/design-system/src/components/Avatar/Avatar.interface.ts
```diff
@@ -29,4 +35,20 @@ export interface AvatarProps
   contentType?: "icon" | "text";
   icon?: TComponentIcon;
   src?: string;
+  statusSign?: ReactNode;
```
monikaracka, line 38:
```suggestion
  statusIcon?: ReactNode;
```

monikaracka (changes requested):

monikaracka (approved):

PR #266, opened 2023-06-07
Ds 522 socials buttons

zuzannanap:


packages/design-system/src/stories/atoms/FacebookButton.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+| DO | <div style="width:20vw">DON’T</div> |
+| ----------- | ----------- |
+| Buttons should: <br/> • Be clearly and accurately labeled.<br/> • Lead with a strong, actionable verb. <br/> • Use established button colors appropriately. For example, only use a red button for an action that’s difficult or impossible to undo. <br/> • Prioritize the most important actions. Too many calls to action can cause confusion and make users unsure of what to do next. <br/> • Be positioned in consistent locations in the interface. ||`;
+
+export const FacebookButton = ({ ...props }: FacebookButtonProps) => (
+  <FacebookButtonComponent {...props} />
+);
+
+export default {
+  title: "🟢 Atoms/SocialsButtons/FacebookButton",
```
monikaracka, line 15: It should be an example, not an atom, because Google, Facebook, or Apple can change their requirements. We can't track this in the Design System, so it's safer to put these buttons as examples.

monikaracka (changes requested):
We also need Apple button.

Add a note about the need to check whether these buttons meet the requirements of Google, Apple or Facebook documentation because it may determine whether the application, especially in mobile applications, will be allowed to be used.
You can add links to documentation.

monikaracka (approved):

PR #276, opened 2023-06-16
Ds 527 radio card

zuzannanap:


kielbasa-elp (changes requested):
Text and Border of RadioCard do not change color to red when it's error. 

Maybe its time to change the behaviour of Radio's and instead of passing error as a Boolean, change it to String as we got in others Inputs ?  

Also, when the Radio is disabled, there is still hover effect on Radio button which is a bit weird. As I see the same behaviour is on the stg

kielbasa-elp (approved):

PR #279, opened 2023-06-19
DS-555 - stepper external steps

monikaracka:


pauzim (approved):
@monikaracka haven't looked at the code, but it seems to work well (checked on storybook). One thing I consider is: could we make these gaps between step item and "line" a little bit smaller? 🤔
Screenshot 2023-06-19 at 18 00 33

PR #301, opened 2023-06-22
Ds 557 RadioCardGroup

zuzannanap:


packages/design-system/src/components/RadioCard/RadioCard.interface.ts
```diff
@@ -1,18 +1,21 @@
...
 
+export type RadioCardSize = "sm" | "md";
+
+export type RadioCardPosition = "left" | "right";
+
 export interface RadioCardProps
   extends Omit<InputHTMLAttributes<HTMLInputElement>, "size">,
     ClassName,
     PropsWithChildren {
-  hasRadio: boolean;
```
kielbasa-elp, line 8: I have few errors in stories because of that

src/stories/examples/RadioCard/AvatarRadioCard.stories.tsx:25:9
for example

kielbasa-elp (approved):
One thing 
There is `isRadioVisible` on the `RadioCardProps` interface  which seems to not work

PR #304, opened 2023-06-25
DS-553 - timepicker

monikaracka:


zuzannanap:
also I think it should be nice if it doesn't start with empty number, so maybe let's make the default hour other than 1:00
(cause i guess loop with 11 and 12 above "1" and 58, 59 above "00" would be hard to be implemented quickly)
image

zuzannanap:
it would be nice to make it possible to click on a visible number to choose it (like here, it would be nice to click on "1" to choose 1 quickly
image

zuzannanap:
would it be possible to make it scrollable by "click and move"? e.g. for somebody who has broken scroll (and btw that's the way i tried to use it automatically for the first time)

zuzannanap:
what is this icon for?
(i can't see it on Figma 🤔)
image

zuzannanap:
i think you can add tokens here - it's a new component so it won't break Zipdo code

packages/design-system/src/components/Timepicker/Timepicker.interface.ts
```diff
@@ -0,0 +1,74 @@
...
+export type MinutesFormatTypes = "quarters" | "tenMinutes" | "minutes";
+export type SecondsFormatTypes = "tenSeconds" | "fiveSeconds" | "seconds";
+export type TimeFormatTypes = "hh:mm" | "hh:mm:ss";
+export enum PeriodTypes {
+  AM = "AM",
+  PM = "PM",
+}
+
+export interface DropdownTimeInterface {
+  hour: number | string;
```
zuzannanap, line 18: hourS (plural, as the rest of periods)

packages/design-system/src/components/Timepicker/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        ? `${time.hour}:${time.minutes}:${time.seconds}`
+        : `${time.hour}:${time.minutes}`;
+
+    if (inputRef.current) {
+      inputRef.current.value = value;
+    }
+    setDropdownOpen(false);
+  };
+
+  const handleKeyDown = (e: KeyboardEvent) => {
```
zuzannanap, line 61: you can use here:
e: React.KeyboardEvent<HTMLInputElement>
and with it in line 105, you don't need @ts-ignore

packages/design-system/src/components/Timepicker/useTimepickerHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+          "10",
+          "15",
+          "20",
+          "25",
+          "30",
+          "35",
+          "40",
+          "45",
+          "50",
+          "55",
```
zuzannanap, line 42: maybe instead of those arrays sth like "createTimePeriods" function that takes gap (like 1, 5, 10) and creates those arrays could be created

zuzannanap (changes requested):

PR #309, opened 2023-06-28
DS-567 - timepicker tokens

monikaracka:


zuzannanap:
gaps between number are a bit bigger than on main (I don't know if it on purpose - looks good, but there's a difference)

zuzannanap:
i think it should be timePicker instead of timepicker everywhere

monikaracka:
> gaps between numbers are a bit bigger than on main (I don't know if it on purpose - looks good, but there's a difference)

This is the difference between 12px and 0.75rem. In Figma, the spacing is 12 px, so I think it can stay that way.

zuzannanap (approved):

PR #320, opened 2023-07-04
DS-476 - tabel filters

monikaracka:


zuzannanap:
TimePicker stories was removed here - now it happens always when you create a branch from main :<

zuzannanap:
filter works only after entering at least 2 characters :<

zuzannanap (approved):

PR #323, opened 2023-07-04
Ds 586 set time picker value on click

zuzannanap:


packages/design-system/src/components/TimePicker/TimePickerDropdown.tsx
```diff
@@ -58,22 +58,26 @@ export const TimePickerDropdown: FC<TimePickerDropdownProps> = ({
...
+    container: React.RefObject<HTMLUListElement>,
+    scrollIndex: number
+  ) => {
+    if (container.current) {
+      container.current.scrollTop = scrollIndex * 40;
     }
+  };
+
+  useEffect(() => {
+    console.log("ok");
```
monikaracka, line 71: Remove console.log

packages/design-system/src/components/TimePicker/index.tsx
```diff
@@ -100,7 +100,12 @@ export const TimePicker: FC<TimePickerProps> = ({
   return (
     <div className="relative">
       <Input
-        onFocus={() => setDropdownOpen(true)}
+        onFocus={() => {
+          setDropdownOpen(true);
+          if (inputRef.current) {
+            inputRef.current.value = "";
+          }
```
monikaracka, line 107: I'm not sure if we should clear input onFocus, because we have Cancel button.

monikaracka (changes requested):

monikaracka (approved):

PR #326, opened 2023-07-05
DS 371: isLoading Button state

zuzannanap:


kielbasa-elp (approved):
Work great. Small thing - loader is not centered

PR #337, opened 2023-07-12
DS-599:Semantic-tabs

zuzannanap:
To increase accessibility, each input should have a label, so they were also added to the timepicker and datepicker (in the end, they would probably come in handy anyway)
I was wondering whether to give the inputs a mandatory id or display the label only if there is an id - the first option seemed to me more readable and intuitive so I used it
Also tags in navbar and herobanner were changed


packages/design-system/src/components/InputCode/index.tsx
```diff
@@ -162,7 +164,10 @@ const InputCode: FC<InputCodeProps> = ({
   return (
     <div className="flex flex-col">
       <input className="peer hidden" disabled={disabled} />
-      <InputText labelHtmlFor={id} text={label} type={getTopTextType({ isDisabled: disabled })} />
+      <InputText
+        text={label}
+        type={getInputTopTextType({ isDisabled: disabled })}
+      />
```
monikaracka, line 170: Why did't you use here `<Label/>`?

packages/design-system/src/components/InputText/InputText.interface.ts
```diff
@@ -3,11 +3,10 @@ import { ReactNode } from "react";
 
 export interface InputTextProps extends ClassName {
   text?: ReactNode;
-  labelHtmlFor?: string;
   type:
-    | "label"
     | "supportingText"
     | "errorMessage"
     | "disabled"
+    | "label"
```
monikaracka, line 10: We shouldn't have component `<Label/>` and label as `variant` in `<InputText/>`.

monikaracka (changes requested):

monikaracka (approved):

PR #338, opened 2023-07-12
add altTextprop to avatar

zuzannanap:


packages/design-system/src/components/Avatar/index.tsx
```diff
@@ -48,8 +49,8 @@ export const Avatar: React.FC<AvatarProps> = ({
         statusNode={statusNode}
         statusPosition={statusPosition}
       >
-        {src ? (
-          <AvatarComponents.Image src={src} />
+        {src && alt ? (
```
monikaracka, line 52: I'm not sure about this solution. If I add `src`, I want to see an image. Maybe we should set the default `alt` eg. "avatar" if `alt` is undefined?

kielbasa-elp: In my opinion "alt" should be optional

packages/design-system/src/components/Avatar/index.tsx
```diff
@@ -9,6 +9,7 @@ export const Avatar: React.FC<AvatarProps> = ({
   caption,
   size = "md",
   src,
+  alt,
```
monikaracka, line 12:
```suggestion
  alt = "avatar",
```

packages/design-system/src/components/Avatar/index.tsx
```diff
@@ -49,7 +50,7 @@ export const Avatar: React.FC<AvatarProps> = ({
         statusPosition={statusPosition}
       >
         {src ? (
-          <AvatarComponents.Image src={src} />
+          <AvatarComponents.Image src={src} alt={alt || "avatar"} />
```
monikaracka, line 53:
```suggestion
          <AvatarComponents.Image src={src} alt={alt} />
```

monikaracka (approved):

PR #343, opened 2023-07-13
DS-602: RangeSlider controls focus

zuzannanap:


monikaracka (changes requested):
It doesn't work good.
1. I can't change the range with my touchpad (or trackpad, I'm not sure about the name 😛 ), only with keyboard.

2. The circle element has a black border when it's focused
Screenshot 2023-07-14 at 08 39 08

3. I have to click outside of the right circle element to focus on it.

https://github.com/elpassion/flounder-2/assets/91883425/83f3755c-b568-4a36-92fa-8be99643fdf2

monikaracka (approved):

PR #346, opened 2023-07-14
DS-611 - accordion tests

monikaracka:


packages/design-system/src/components/Accordion/__tests__/Accordion.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+    },
+  ];
+
+  const mountComponent = (items = itemsMock) =>
+    render(<Accordion items={items} variant="bordered" />);
+
+  it("should have visibled accordion body with appropriate id after button click", async () => {
+    const view = mountComponent();
+
+    const firstAccordionBody = screen.getByTestId("accordion-1");
```
zuzannanap, line 46: it would be nice if we have a class like "AccordionTestComponent"with methods like "mountComponent", "getFirstAccordion", "getFirstButton", "clickFirstButton" etc. - it makes tests much more readable

packages/design-system/src/components/Accordion/__tests__/Accordion.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+      title: "Accordion 3",
+      description: "Accordion 3 description",
+    },
+  ];
+
+  const mountComponent = (items = itemsMock) =>
+    render(<Accordion items={items} variant="bordered" />);
+
+  it("should have visibled accordion body with appropriate id after button click", async () => {
+    const view = mountComponent();
```
zuzannanap, line 44: do we need this "view" variable? woudn't just "mountComponent()" be enough?

packages/design-system/src/components/Accordion/Accordion.interface.ts
```diff
@@ -35,4 +35,5 @@ export interface AccordionBodyProps
   extends Pick<AccordionItemProps, "description">,
     Pick<AccordionProps, "iconPosition"> {
   isExpanded: boolean;
+  id: number;
```
zuzannanap, line 38: adding id just for test-id is not the best option
you can search for it by e.g. getByText - it's a more recommended practice (it would be more similar to user experience)

zuzannanap (changes requested):

packages/design-system/src/components/Accordion/Accordion.interface.ts
monikaracka: I can't use getByText because it can only find the child that way.

zuzannanap: ahh right

packages/design-system/src/components/Accordion/__tests__/Accordion.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <AccordionComponent
+      {...props}
+      onClick={(newId) => setExpandedId(newId)}
+      expandedId={expandedId}
+    />
+  );
+};
+class AccordionObject {
```
zuzannanap, line 61: maybe "AccordionTestComponent"?

monikaracka: But it's not a component. It's an object like in the pattern's name.

packages/design-system/src/components/Accordion/__tests__/Accordion.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  get accordionBodys() {
+    return screen.getAllByTestId("accordion-body");
+  }
+
+  get accordionButtons() {
+    return screen.getAllByRole("button");
+  }
+
+  async clickAt(button: HTMLElement) {
```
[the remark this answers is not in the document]
monikaracka: Why do you think it's a better way to write this function?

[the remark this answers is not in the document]
zuzannanap: ok, you're right

packages/design-system/src/components/Accordion/__tests__/Accordion.spec.tsx
zuzannanap: sorry, it should be comment to line 27 and relate to:
```
export const Accordion = ({ ...props }: AccordionComponentProps) => {
  const [expandedId, setExpandedId] = useState<number>();

  return (
    <AccordionComponent
      {...props}
      onClick={(newId) => setExpandedId(newId)}
      expandedId={expandedId}
    />
  );
};
```

packages/design-system/src/components/Accordion/__tests__/Accordion.screen-object.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  get accordionButtons() {
+    return screen.getAllByRole("button");
+  }
+
+  async clickAt(button: HTMLElement) {
+    if (!button) throw new Error(`Button not found`);
+
+    return fireEvent.click(button);
```
zuzannanap, line 66: it's recommended to use user-event when it's possible
https://testing-library.com/docs/user-event/intro/#:~:text=Differences%20from%20fireEvent%20%E2%80%8B,additional%20checks%20along%20the%20way.

packages/design-system/src/components/Accordion/__tests__/Accordion.screen-object.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  get accordionButtons() {
+    return screen.getAllByRole("button");
+  }
+
+  clickAt(button: HTMLElement) {
+    if (!button) throw new Error(`Button not found`);
+
+    return userEvent.click(button);
```
zuzannanap, line 69: you can write just button.click(), but whatever you like :)

zuzannanap (approved):

PR #347, opened 2023-07-14
DS-612 - anchor test

monikaracka:


packages/design-system/src/components/Anchor/__tests__/Anchor.spec.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+describe("Anchor", () => {
+  const mockTextLink = "Click me";
+  const mockLinkHref = "https://www.google.com/";
+
+  const mountComponent = (text = mockTextLink, link = mockLinkHref) =>
+    render(<Anchor text={text} link={link} />);
+
+  it('should be a link that have href value to "https://www.google.com"', () => {
+    const view = mountComponent();
```
kielbasa-elp, line 12: let's just execute the mountComponent without creating view variable

kielbasa-elp (approved):

zuzannanap:
same as with accordion - nice to have a class with mount method
also you can check if the element can be clicked

PR #349, opened 2023-07-17
Ds 610 aria attributes part1

zuzannanap:


packages/design-system/src/components/FAB/FAB.interface.ts
```diff
@@ -1,12 +1,18 @@
 import { BaseComponentSizes, ClassName, TComponentIcon } from "@helpers/types";
 
+export type label =
```
kielbasa-elp, line 3: It's type so create Label instead of label

packages/design-system/src/components/Input/index.tsx
```diff
@@ -46,6 +46,8 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
           suffixIcon={suffixIcon}
           value={value}
           ref={ref}
+          ariaLabel={label.ariaLabel || label.ariaLabel}
```
kielbasa-elp, line 49: And i don't like that we override default html aria property in our components. Our component extends `HTMLInputElement` so it's good to just spread rest properties. So we could use it like this:
<Input aria-label="label" aria-describedby='label' />
instead of
<Input ariaLabel="label" ariaDescribedBy='label' />

PR #355, opened 2023-07-17
Ds 610 aria attributes part2

zuzannanap:


packages/design-system/src/components/Dropdown/DropdownComponents.tsx
```diff
@@ -43,6 +43,7 @@ export const DropdownComponents = {
     isSelected,
     onSelect,
     id,
+    toggleLabel,
```
monikaracka, line 46: I'm not sure that we need toggleLabel in dropdown option, and we can't pass it in e.g. SelectDropdown.

packages/design-system/src/components/RangeSlider/RangeSlider.interface.ts
```diff
@@ -2,6 +2,10 @@ export interface RangeSliderProps {
   step: number;
   value: IRange;
   labelPosition: TLabelPosition;
+  slidersAriaLabels: {
+    minValueSliderAriaLabel: string;
+    maxValueSliderAriaLabel: string;
+  };
```
monikaracka, line 8:
```suggestion
    minValueSliderAriaLabel?: string;
    maxValueSliderAriaLabel?: string;
```
I agree with @kielbasa-elp we shouldn't make ariaLabels obligatory.

packages/design-system/src/components/Toggle/index.tsx
```diff
@@ -42,7 +41,7 @@ export const Toggle = forwardRef<HTMLInputElement, ToggleProps>(
             disabled={disabled}
             aria-disabled={disabled}
             aria-describedby="help"
-            aria-label={ariaLabel}
+            aria-label={label.ariaLabel || label.visibleLabel}
```
monikaracka, line 44:
```suggestion
            aria-label={props["aria-label"] || "toggle"}
```
or
```suggestion
            aria-label={props["aria-label"] || label}
```

packages/design-system/src/components/Toggle/Toggle.interface.ts
```diff
@@ -1,16 +1,19 @@
...
+export type label =
+  | { visibleLabel: string; ariaLabel?: string }
+  | { visibleLabel?: string; ariaLabel: string };
+
 export interface ToggleProps
   extends Omit<InputHTMLAttributes<HTMLInputElement>, "size" | "onChange">,
     ClassName {
   size: "sm" | "md";
-  labelText?: string;
+  label: label;
```
monikaracka, line 12:
```suggestion
  label: string;
```
Label, which is an object, is not very intuitive.

monikaracka (changes requested):

packages/design-system/src/components/Dropdown/DropdownComponents.tsx
zuzannanap: but we don't have to - it's optional

monikaracka (changes requested):
Update Toggle components in table utils

kielbasa-elp (approved):

packages/design-system/src/components/Stepper/Stepper.interface.ts
```diff
@@ -13,6 +13,7 @@ interface StepInterface {
   icon?: TComponentIcon;
   stepTitle?: string;
   stepDescription?: string;
+  stepButtonAriaLabel: string;
```
kielbasa-elp: maybe we can extend stepProp by htmlButton props and pick aria-label etc?

monikaracka (approved):

packages/design-system/src/components/SidePanel/index.tsx
```diff
@@ -26,18 +22,20 @@ const SidePanel: FC<SidePanelProps> = ({
...
         className={classNames(
-          "pb-sidePanel-paddingHeading mb-sidePanel-marginBase flex gap-x-6 border-b",
+          "mb-sidePanel-marginBase flex gap-x-6 border-b pb-sidePanel-paddingHeading",
           iconPositionVariants[iconPosition]
         )}
       >
         <IconButton
-          icon={icon}
+          customIcon={iconButton.customIcon}
+          iconName={iconButton.iconName}
```
kielbasa-elp, line 34: Now i am thinking maybe its better to leave `icon` here instead of customIcon. It's a bit strange to use `IconButton` and pass a customIcon

packages/design-system/src/components/Stepper/Stepper.interface.ts
zuzannanap: upsi, sure :)

packages/design-system/src/components/SidePanel/index.tsx
zuzannanap: ok ;(

PR #357, opened 2023-07-18
Ds 622 accordion body resizing

zuzannanap:


monikaracka (commented):
It looks like something that we can reuse in other components. Creating a custom hook sounds like a good idea in this case. What do you think?

packages/design-system/src/components/Accordion/AccordionComponents.tsx
```diff
@@ -7,28 +7,51 @@ import {
...
+    isExpanded,
+  }: AccordionBodyProps) => {
     const iconVariants = {
       left: "pl-accordion-paddingIconLeft",
       right: "pl-accordion-paddingIconRight",
     };
 
     const [contentHeight, setContentHeight] = useState<number>();
     const contentRef = useRef<HTMLParagraphElement>(null);
+    const windowWidth = window.innerWidth;
```
kielbasa-elp, line 22: Do not use window directly in component body like here because it will cause runtime error if not defined

kielbasa-elp (changes requested):
Do not use

kielbasa-elp (commented):
There is useWindowSize hook already in utils folder. Maybe you can use it?

kielbasa-elp (approved):

PR #361, opened 2023-07-18
DS-613 - avatar test

monikaracka:


packages/design-system/src/components/Avatar/__tests__/Avatar.spec.tsx
```diff
@@ -0,0 +1,51 @@
+import { render, screen } from "@testing-library/react";
+import { getImageUrl } from "stories/utils";
+import Avatar from "../index";
+import { AvatarProps } from "../Avatar.interface";
+
+describe("Avatar", () => {
+  it("should be rendered", () => {
+    const { avatarContainer } = new AvatarObject().render();
+
+    expect(avatarContainer).toBeInTheDocument();
```
zuzannanap, line 10: i'm not sure if we need this test - if it fails, we'd know it from other tests

packages/design-system/src/components/Avatar/__tests__/Avatar.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+  it("should have name text", () => {
+    const { avatarText, avatarImage } = new AvatarObject().render();
+
+    expect(avatarImage).toBeNull();
+    expect(avatarText).toBeInTheDocument();
```
zuzannanap, line 16: it's the best when a single test has only one assertion - i think we don't need checking if avatar image is not there when we don't have any "src"
but it might be tested in a separate test too

packages/design-system/src/components/Avatar/__tests__/Avatar.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+  it("should have image not name text", () => {
+    const { avatarImage } = new AvatarObject().render({
+      src: getImageUrl("/red.png"),
+    });
+
+    expect(avatarImage).toBeInTheDocument();
+  });
+});
+
+class AvatarObject {
```
zuzannanap, line 27: screen-object.tsc, same as before

packages/design-system/src/components/Avatar/index.tsx
```diff
@@ -41,7 +41,7 @@ export const Avatar: React.FC<AvatarProps> = ({
   };
 
   return (
-    <div className="flex items-center">
+    <div className="flex items-center" data-testid="avatar">
```
zuzannanap, line 44: i'm not sure if hardcoding data-testid is a good option, but as i said - i'm not sure :D

zuzannanap (approved):

PR #366, opened 2023-07-19
Ds 638 sidebar tests

kielbasa-elp:


packages/design-system/src/components/Sidebar/index.tsx
```diff
@@ -20,11 +20,12 @@ const Sidebar = (props: SidebarProps) => {
     breakpoints,
     defaultCollapsed = false,
     collapseButton = true,
+    collapsed: parentCollapsed,
```
zuzannanap, line 23: why didn't you change it in the interface?

packages/design-system/src/components/Sidebar/index.tsx
```diff
@@ -123,6 +135,7 @@ function ToggleButton({ renderButton = true, className }: CollapseButtonProps) {
         size="sm"
         className="!border-sidebar-base-toggleButton !bg-sidebar-base-toggleButton"
         onClick={toggleCollapse}
+        aria-label={isCollapsed ? "Expand sidebar" : "Collapse sidebar"}
```
zuzannanap, line 138: it's ok, but only for english version

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const { container } = render(
+      <TestSidebar id="sidebar" testId="sidebar" {...props} />
+    );
+    this._container = container;
+
+    return this;
+  }
+
+  get toggleButton() {
+    return getByRole(this._container, "button", { name: /collapse sidebar/i });
```
zuzannanap, line 34: we (Monika and me) use screen.getByRole - could you use it to make tests more consistent? Or are you doing it this way for a special purpose?

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  getSidebar(id = "sidebar") {
+    return getByTestId(this._container, id);
+  }
+
+  toggleSidebar() {
+    return act(async () => {
+      await userEvent.click(this.toggleButton);
```
zuzannanap, line 43: this.toggleButton.click() will work the same and will be consistent with other tests

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    await sidebarObject.toggleSidebar();
+
+    expect(sidebarObject.getSidebar()).toHaveAttribute(
+      "aria-expanded",
+      "false"
+    );
+  });
+});
+
+class SidebarObject {
```
zuzannanap, line 21: please move it to Sidebar.screen-object.tsx

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+describe(SidebarComponent.name, () => {
+  it("should toggle sidebar", async () => {
+    const sidebarObject = new SidebarObject().render();
+
+    expect(sidebarObject.getSidebar()).toHaveAttribute("aria-expanded", "true");
+
+    await sidebarObject.toggleSidebar();
+
+    expect(sidebarObject.getSidebar()).toHaveAttribute(
```
zuzannanap, line 14: wouldn't it be better to divide it for 2 separate tests? (only one assertion for a test is a good practice)

zuzannanap (changes requested):

zuzannanap:
please change test.tsx for spec.tsx to be consistent with other tests :)

packages/design-system/src/components/Sidebar/index.tsx
kielbasa-elp: Why should I ? It's much better just to use e.g. `<Sidebar collapsed={true} />` than` <Sidebar parentCollapsed={true}/>` i think. Dev doesn't need to know what is going inside component

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
kielbasa-elp: This approach limit the search scope of the element and get button only from a given container, not from the entire body. It can be useful in some cases. 

But i got your point and sure, i will change it

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
kielbasa-elp: sure, but i am curious why "screen-object" ?

packages/design-system/src/components/Sidebar/index.tsx
kielbasa-elp: Dev has possibility to pass custom ToggleButton with translated labels etc. I am not sure if it is necessary to add additional properties just for this case

packages/design-system/src/components/Sidebar/__tests__/Sidebar.test.tsx
kielbasa-elp: I don't know 😄 i mean, i wanted to test collapse/expand functionality. I can create two separate tests and pass different initial state but, i think it won't be the same

packages/design-system/src/components/Sidebar/index.tsx
zuzannanap: okok

packages/design-system/src/components/Sidebar/index.tsx
zuzannanap: yeah right

monikaracka (approved):

PR #368, opened 2023-07-19
DS 635: Breadcrumbs unit tests

zuzannanap:


packages/design-system/src/components/Breadcrumbs/__tests__/Breadcrumbs.spec.tsx
```diff
@@ -0,0 +1,28 @@
+import Breadcrumbs from "..";
+import { BreadcrumbsObject, mockIcon } from "./Breadcrumbs.screen-object";
+
+describe(Breadcrumbs.name, () => {
+  it("displays only 3 breadcrumbs when showAll is set to false", () => {
+    const screen = new BreadcrumbsObject().render();
+
+    expect(screen.breadcrumbs).toHaveLength(3);
+  });
+  it("displays only 3 breadcrumbs when showAll is set to true", () => {
```
kielbasa-elp, line 10: wrong test name here i guess

packages/design-system/src/components/Breadcrumbs/__tests__/Breadcrumbs.screen-object.tsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+  get firstLinkHref() {
+    return this.breadcrumbs[0].getAttribute("href");
+  }
+
+  render(props?: Partial<BreadcrumbsProps>) {
+    const { container } = render(
+      <Breadcrumbs
+        breadcrumbs={breadCrumbsMock}
+        showAll={props?.showAll || false}
```
kielbasa-elp, line 60: Maybe it's good idea to set default value in component e.g. false?

zuzannanap: oh, it was, i missed it

zuzannanap: but it was also required, so i will change it

kielbasa-elp (approved):

PR #369, opened 2023-07-19
Ds 641 carousel tests

kielbasa-elp:


packages/design-system/src/components/Carousel/__tests__/Carousel.screen-object.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  get slides() {
+    return screen.getAllByRole("img", { name: /slide/i });
+  }
+
+  getDots(name = /show slide/i) {
+    return screen.queryAllByRole("button", { name });
+  }
+
+  getPrevButton(name = /Previous slide/i) {
```
zuzannanap, line 24: does it make sense to create a param in this method instead of just creating a getter without a param?

packages/design-system/src/components/Carousel/__tests__/Carousel.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+describe(Carousel.name, () => {
+  it("should render correct amount of slides", async () => {
+    const carouselObject = new CarouselObject().render({
+      slides: generateSlides(5),
+    });
+
+    expect(carouselObject.slides).toHaveLength(5);
+  });
+
+  it("should not render navigation", async () => {
```
zuzannanap, line 13: i'd  add "when ..." to this title - it won't be clear in the console

packages/design-system/src/components/Carousel/__tests__/Carousel.screen-object.tsx
kielbasa-elp: Yeap, because dev can pass custom button with custom name like in the tests above

kielbasa-elp:
@monikaracka when it comes to tests indicators - this is the same case as bacToTop Button. I am not sure if it is testable in jest envviroiment

monikaracka (approved):

PR #371, opened 2023-07-19
Ds 462 checkbox unit tests

zuzannanap:


packages/design-system/src/components/Checkbox/__tests__/Checkbox.screen-object.tsx
```diff
@@ -0,0 +1,35 @@
+import { CheckboxProps } from "@components/Checkbox/Checkbox.interface";
+import { render, screen } from "@testing-library/react";
+import Checkbox from "..";
+
+export const mockClick = jest.fn();
```
kielbasa-elp, line 5: If it's not used, just remove it

kielbasa-elp (approved):

PR #379, opened 2023-07-21
DS-624 - back to top button tests

monikaracka:


packages/design-system/src/components/BackToTopButton/__tests__/useScrolledButtonVisibility.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+  buttonRightPxPosition: 0,
+};
+
+describe("BackToTopButton", () => {
+  it("shouldn't be visible after render", async () => {
+    const { result } = renderHook(() =>
+      useScrolledButtonVisibility(inititialValues)
+    );
+
+    expect(result.current.buttonVisible).toBe(false);
```
zuzannanap, line 16: maybe try to use .not.toBeVisible() ?

packages/design-system/src/components/BackToTopButton/__tests__/BackToTopButton.spec.tsx
```diff
@@ -0,0 +1,9 @@
+import { BackToTopButtonObject } from "./BackToTopButton.screen-object";
+
+describe("BackToTopButton", () => {
+  it("shouldn't be visible after render", async () => {
+    const { container } = new BackToTopButtonObject().render();
+
+    expect(container).toHaveClass("hidden");
```
zuzannanap, line 7: maybe you could also check if it's visible when it should? :)

monikaracka: No, because I can't imitate scrolling because of the test environment in jest.

packages/design-system/src/components/BackToTopButton/__tests__/useScrolledButtonVisibility.spec.tsx
monikaracka: But it's not a component. It's the value of `buttonVisible`, witch is boolean.

zuzannanap (approved):

PR #380, opened 2023-07-21
Add featuredCard tests

kielbasa-elp:


packages/design-system/src/components/FeaturedCard/__tests__/FeaturedCard.screen-object.tsx
```diff
@@ -0,0 +1,45 @@
...
+    );
+
+    return this;
+  }
+
+  get collapsedButton() {
+    return screen.getByRole("button", { name: /featured card/i });
+  }
+
+  clickCollapsedCard() {
```
zuzannanap, line 26: as i said in carousel CR: you don't have to wrap in in act when using eserEvent instead of fireEvent https://github.com/testing-library/user-event/discussions/906

packages/design-system/src/components/FeaturedCard/__tests__/FeaturedCard.screen-object.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    return this;
+  }
+
+  get closeButton() {
+    return screen.getByRole("button", { name: /close/i });
+  }
+
+  clickCloseButton() {
+    act(() => {
```
zuzannanap, line 39: same here

packages/design-system/src/components/FeaturedCard/index.tsx
```diff
@@ -43,7 +45,9 @@ export const FeaturedCard: React.FC<FeaturedCardProps> = (props) => {
     return (
       <IconButton
         icon={icon}
+        onClick={onCollapsedClick}
         variant="basic"
+        aria-label="Featured card"
```
zuzannanap, line 50: "open" will make more sense, but anyway, it'll be covered in one of aria-PRs

PR #383, opened 2023-07-24
Add modal tests

kielbasa-elp:


packages/design-system/src/components/Modal/__tests__/Modal.screen-object.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return this;
+  }
+
+  openModal() {
+    const button = screen.getByRole("button", { name: /open modal/i });
+
+    act(() => {
+      button.click();
+    });
+
```
kielbasa-elp, line 21: This click manipulates state so it have to be wrapped by act

packages/design-system/src/components/Modal/__tests__/Modal.screen-object.tsx
```diff
@@ -0,0 +1,54 @@
...
+  get modal() {
+    return screen.queryByRole("dialog");
+  }
+
+  closeModal() {
+    const overlay = document.querySelector(".ReactModal__Overlay");
+    if (!overlay) throw new Error("Modal not opened");
+
+    act(() => {
```
zuzannanap, line 33: https://plainenglish.io/blog/you-probably-dont-need-act-in-your-react-tests-2a0bcd2ad65c

zuzannanap: we can use userEvent.click(element) without wrapping in act

packages/design-system/src/components/Modal/__tests__/Modal.screen-object.tsx
zuzannanap: not when you use userEvent

packages/design-system/src/components/Modal/__tests__/Modal.screen-object.tsx
```diff
@@ -0,0 +1,54 @@
...
+      (overlay as HTMLElement).click();
+    });
+
+    return this;
+  }
+}
+
+const TestModal = ({ ...props }: Partial<ModalProps>) => {
```
zuzannanap, line 41: TestModalComponents - to be consistent

packages/design-system/src/components/Modal/__tests__/Modal.screen-object.tsx
zuzannanap: i think it's more readable

zuzannanap (approved):

packages/design-system/src/components/Modal/__tests__/Modal.screen-object.tsx
zuzannanap, line 41: TestModalComponent to be consistent

zuzannanap (changes requested):

PR #385, opened 2023-07-24
DS-647 - FAB tests

monikaracka:


packages/design-system/src/components/FAB/__tests__/FAB.screen-object.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const onClickMock = jest.fn();
+
+const items = [
+  { label: "Inbox", onClick: onClickMock, icon: <BellSvg /> },
+  { label: "Profile", onClick: onClickMock, icon: <UserSvg /> },
+  { label: "Settings", onClick: onClickMock, icon: <CogSvg /> },
+];
+
+export class FABObject {
+  private _container!: HTMLElement;
```
kielbasa-elp, line 18: Because we decided to use screen, I think the _container is no longer needed

kielbasa-elp (approved):

PR #388, opened 2023-07-25
fix/DS-657 - input icon button type

monikaracka:


packages/design-system/src/components/Input/InputComponents.tsx
```diff
@@ -242,14 +246,20 @@ const InputComponents = {
...
     return (
       <button
         className={classNames(
           "absolute right-3.5 row-start-2 row-end-3 flex h-full w-4 cursor-pointer items-center text-input-base-suffixIcon",
           "peer-disabled:text-input-base-disabled",
           positionStyle
         )}
-        onClick={onIconClick}
+        onClick={handleIconClick}
+        type="button"
```
kielbasa-elp, line 262: i am not sure, but I think "button" is a default type

kielbasa-elp (approved):

packages/design-system/src/components/Input/InputComponents.tsx
monikaracka: Probably yes, but if we don't pass it form doesn't work correctly with the keyboard. I had this bug in the login form for another project.

PR #396, opened 2023-07-25
Ds 659 IconButton unit tests

zuzannanap:


packages/design-system/src/components/IconButton/__tests__/IconButton.screen-object.tsx
```diff
@@ -0,0 +1,22 @@
...
+import IconButton from "..";
+import { IconButtonProps } from "../IconButton.interface";
+
+export const mockIcon = <Icon iconName="search" />;
+export class IconButtonObject {
+  get iconButton() {
+    return screen.getByRole("button");
+  }
+  clickIconButton() {
+    act(() => this.iconButton.click());
```
monikaracka, line 12: use `userEvent` to click - ['userEvent or some event functions are already wrapped in act from testing-library/react, therefore you don’t need to use act() actually.'](https://flyingsquirrel.medium.com/when-should-i-use-act-in-react-testing-library-d7dd22a3340e)

packages/design-system/src/components/IconButton/__tests__/IconButton.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const onClick = jest.fn();
+    const screen = new IconButtonObject().render({
+      onClick,
+      disabled: true,
+    });
+
+    screen.clickIconButton();
+    expect(onClick).not.toHaveBeenCalled();
+  });
+  it("dislays icon", () => {
```
monikaracka, line 24:
```suggestion
  it("displays icon", () => {
```
The same mistake is in button test 😉

monikaracka (approved):

PR #408, opened 2023-07-27
DS 665 Navbar unit tests

zuzannanap:


packages/design-system/src/components/Navbar/__tests__/Navbar.screen-object.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const mockMenuClick = jest.fn();
+export class NavbarObject {
+  get navbar() {
+    return screen.getByTestId("navbar");
+  }
+  get openMenuButton() {
+    return screen.getByRole("button", { name: "open menu" });
+  }
+  get closeMenuButton() {
+    return screen.getByRole("button", { name: "open menu" });
```
monikaracka, line 15: You didn't change the name of the close button.

packages/design-system/src/components/Navbar/__tests__/Navbar.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+    expect(children).toBeInTheDocument();
+  });
+  it("allows to Click menu button", () => {
+    new NavbarObject().render({
+      children: <Button text="Button text" />,
+    });
+
+    const children = screen.getByRole("button", { name: "Button text" });
+
+    expect(children).toBeInTheDocument();
```
monikaracka, line 28: This is the same test as the previous one.  The description was changed but not the test.

monikaracka (approved):

PR #415, opened 2023-07-28
DS-666 InputText unit tests

zuzannanap:


packages/design-system/src/components/InputText/__tests__/InputText.screen-object.tsx
```diff
@@ -0,0 +1,25 @@
...
+import InputText from "..";
+import { InputTextProps } from "../InputText.interface";
+
+export const mockIcon = <Icon iconName="search" />;
+export class InputTextObject {
+  get text() {
+    return screen.getByText("Text");
+  }
+  get icons() {
+    return screen.queryAllByRole("img");
```
monikaracka, line 12: Should it be here?

PR #421, opened 2023-07-31
DS-661 - indicator tests

monikaracka:


packages/design-system/src/components/Indicator/__tests__/Indicator.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+    expect(dot).not.toBeInTheDocument();
+  });
+
+  it("should have red dot in error type", () => {
+    const { dot } = new IndicatorObject().render({ type: "error" });
+
+    expect(dot).toHaveClass("bg-indicator-error-dot");
+  });
+
+  it("should have blue background in processing type", () => {
```
zuzannanap, line 43: maybe not blue, but just "processing color background"? but this way is also ok :)

packages/design-system/src/components/Indicator/__tests__/Indicator.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const { wrapper } = new IndicatorObject().render({
+      variant: "count",
+      type: "processing",
+    });
+
+    expect(wrapper).toHaveClass("bg-indicator-processing-dot");
+  });
+
+  it("should have green background in success type", () => {
```
zuzannanap, line 52: same

packages/design-system/src/components/Indicator/__tests__/Indicator.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const { wrapper } = new IndicatorObject().render({
+      variant: "badge",
+      type: "success",
+    });
+
+    expect(wrapper).toHaveClass("bg-indicator-success");
```
zuzannanap, line 58: same

zuzannanap (approved):

PR #427, opened 2023-08-01
DS 682: LinearProgressBar and CircularProgressBar unit tests

zuzannanap:


packages/design-system/src/components/ProgressBar/__tests__/LinearProgressBar.screen-object.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export class LinearProgressBarObject {
+  get progressBar() {
+    return screen.getByTestId("linearProgressBar");
+  }
+  get label() {
+    return screen.getByText("60%");
+  }
+
+  get wholeLine() {
```
monikaracka, line 13: What about `backgroundLine` ?

monikaracka (approved):

packages/design-system/src/components/ProgressBar/__tests__/LinearProgressBar.screen-object.tsx
zuzannanap: ok :) i think it's better

PR #429, opened 2023-08-02
DS-685 - radio card group

monikaracka:


packages/design-system/src/components/RadioCardGroup/__tests__/RadioCardGroup.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const view = new RadioCardGroupObject().render();
+
+    view.assertCheckedRadio();
+  });
+
+  it("should check radio after card click", async () => {
+    const view = new RadioCardGroupObject().render();
+
+    await view.clickCard(0);
+    expect(view.radios[0]).toBeChecked();
```
zuzannanap, line 56: why don't you use here assertCheckedRadio?

packages/design-system/src/components/RadioCardGroup/__tests__/RadioCardGroup.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  it("should check only one radio card after click", async () => {
+    const view = new RadioCardGroupObject().render();
+
+    await view.clickCard(0);
+    expect(view.radios[0]).toBeChecked();
```
zuzannanap, line 63: ad here too?

zuzannanap (approved):

packages/design-system/src/components/RadioCardGroup/__tests__/RadioCardGroup.spec.tsx
monikaracka: In this test case, I want to check only one radio after click.

packages/design-system/src/components/RadioCardGroup/__tests__/RadioCardGroup.spec.tsx
monikaracka: Here I changed it 😄

PR #437, opened 2023-08-04
DS-681 - pagination tests

monikaracka:


packages/design-system/src/components/Pagination/__tests__/Pagination.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    expect(dotButton).toBeInTheDocument();
+  });
+
+  it("should display seven buttons", () => {
+    const { buttons } = new PaginationObject().render();
+
+    expect(buttons).toHaveLength(7);
+  });
+
+  it("should click previous button", async () => {
```
zuzannanap, line 34: should allow to click?

packages/design-system/src/components/Pagination/__tests__/Pagination.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const view = new PaginationObject().render();
+
+    view.clickButton(view.previousButton);
+    await waitFor(() => expect(onPreviousMock).toHaveBeenCalled());
+  });
+
+  it("should click next button", async () => {
```
zuzannanap, line 41: same

packages/design-system/src/components/Pagination/__tests__/Pagination.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const view = new PaginationObject().render();
+
+    view.clickButton(view.nextButton);
+    await waitFor(() => expect(onNextMock).toHaveBeenCalled());
+  });
+
+  it("should click page button", async () => {
```
zuzannanap, line 48: same

packages/design-system/src/components/Pagination/__tests__/Pagination.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  it("should have disabled next button when nextButtonDisabled props is true", async () => {
+    const { nextButton } = new PaginationObject().render({
+      nextButtonDisabled: true,
+    });
+
+    expect(nextButton).toBeDisabled();
+  });
+
+  it("shouldn't display dot button when number of pages is lower than five", async () => {
```
zuzannanap, line 71: not related to this pr: i think it shouldn't display dots button also when number of pages is 5 - it should just display all the pages, don't you think?

packages/design-system/src/components/Pagination/__tests__/Pagination.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it("should click page button", async () => {
+    const view = new PaginationObject().render();
+
+    view.clickButton(view.buttons[2]);
+    await waitFor(() => expect(onPageMock).toHaveBeenCalled());
+  });
+
+  it("should have disabled previous button when prevButtonDisabled props is true", async () => {
```
zuzannanap, line 55: not related to pr, but can be fixed here: it should be also disabled when current page is one (and next button should be disabled when curret page is equal to number of pages)

zuzannanap (approved):

PR #438, opened 2023-08-04
DS-687 - rating tests

monikaracka:


packages/design-system/src/components/Rating/__tests__/Rating.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    expect(halfRatingStars).toHaveLength(5);
+  });
+
+  it("should call half stars buttons after click", async () => {
+    const view = new RatingObject().render({ precision: 0.5 });
+
+    view.clickStars(view.halfRatingStars);
+    await waitFor(() => expect(onClickMock).toHaveBeenCalledTimes(5));
+  });
```
zuzannanap, line 47: i think it should be nice to test it by creating RatingTestComponent with value state cause onClick function has different arguments (value and value-0.5) in code and it'll be good to test it

zuzannanap (approved):

PR #448, opened 2023-08-07
DS 690-search-unit-tests

zuzannanap:


packages/design-system/src/svgs/SearchSvg.tsx
```diff
@@ -8,6 +8,9 @@ const SearchSvg: FC<ClassName> = ({ className }) => {
       fill="none"
       xmlns="http://www.w3.org/2000/svg"
       className={className}
+      aria-hidden
+      role="img"
+      data-testid="searchSvg"
```
monikaracka, line 13: I don't think we should add test id to icons.

packages/design-system/src/components/Search/__tests__/Search.screen-object.tsx
```diff
@@ -0,0 +1,28 @@
...
+import Search from "..";
+import { SearchProps } from "../Search.interface";
+
+export const mockErrorMessage = "Error";
+export class SearchObject {
+  get input() {
+    return screen.getByPlaceholderText("Placeholder");
+  }
+  get icon() {
+    return screen.getByTestId("searchSvg");
```
monikaracka, line 12:
```suggestion
    return screen.getByRole("img", { hidden: true });
```

monikaracka (approved):

packages/design-system/src/svgs/SearchSvg.tsx
zuzannanap: hmm weird i don't have it on local branch 🙃

zuzannanap: oh ok, i haven't see you commented it before i removed it, nvm :D

PR #451, opened 2023-08-08
Ds 702 TimePicker unit tests

zuzannanap:


packages/design-system/src/components/TimePicker/__tests__/TimePicker.screen-object.tsx
```diff
@@ -0,0 +1,99 @@
...
+};
+export class TimePickerObject {
+  get input(): HTMLInputElement {
+    return screen.getByPlaceholderText("Placeholder");
+  }
+  get supportingText() {
+    return screen.getByText("Supporting text");
+  }
+  get label() {
+    return screen.getByText("Supporting text");
```
monikaracka, line 45:
```suggestion
    return screen.getByText("Label");
```

packages/design-system/src/components/TimePicker/__tests__/TimePicker.screen-object.tsx
```diff
@@ -0,0 +1,99 @@
...
+    userEvent.click(this.timePart(timePart, time));
+  }
+  async typeInInput(content: string) {
+    await userEvent.type(this.input, content);
+  }
+  async focusInput() {
+    await userEvent.click(this.input);
+  }
+  async clickCloseButton() {
+    const button = screen.getByRole("button", { name: "Cancel" });
```
monikaracka, line 78: Use a function to get buttons, e.g.:
```
getButton( name: string ) {
   screen.getByRole("button", { name }) 
}
```

Or separate the Cancel button and OK button. But try to do only one thing in one function.

packages/design-system/src/components/TimePicker/__tests__/TimePicker.screen-object.tsx
```diff
@@ -0,0 +1,99 @@
...
+    await userEvent.click(button);
+  }
+  async clickOKButton() {
+    const button = screen.getByRole("button", { name: "OK" });
```
monikaracka, line 82: same here

packages/design-system/src/components/TimePicker/__tests__/TimePicker.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+  });
+  it("doesn't allow to type in input when disabled", () => {
+    const view = new TimePickerObject().render({ disabled: true });
+    view.typeInInput("11:11");
+
+    expect(view.input).toHaveValue("");
+  });
+  it("doesn't allow to type letters in input", () => {
+    const view = new TimePickerObject().render();
+    view.typeInInput("abc");
```
monikaracka, line 35: add space before this line and I don't see `expect` here

packages/design-system/src/components/TimePicker/__tests__/TimePicker.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+  });
+  it("displays period buttons", async () => {
+    const { AMButton, PMButton } = new TimePickerObject().render({
+      withPeriodButtons: true,
+    });
+
+    expect(AMButton).toBeInTheDocument();
+    expect(PMButton).toBeInTheDocument();
+  });
+  it("douesn't displays period buttons when withPeriodButtons prop is false", async () => {
```
monikaracka, line 77: fix typo 😉

packages/design-system/src/components/TimePicker/__tests__/TimePicker.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    expect(AMButton).not.toBeInTheDocument();
+    expect(PMButton).not.toBeInTheDocument();
+  });
+  it("allows to change hours format for 24 hours", async () => {
+    const view = new TimePickerObject().render({ hoursFormat: 24 });
+
+    view.focusInput();
+    const hour23 = view.timePart("hours", "23");
```
monikaracka, line 96: What about `expectedHours`?

packages/design-system/src/components/TimePicker/__tests__/TimePicker.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    expect(hour23).toBeInTheDocument();
+  });
+  it("allows to change minutes format for minutes", async () => {
+    const view = new TimePickerObject().render({
+      minutesFormat: "minutes",
+    });
+
+    view.focusInput();
+    const minutes59 = view.timePart("minutes", "59");
```
monikaracka, line 106: What about `expectedMinutes`?

packages/design-system/src/components/TimePicker/__tests__/TimePicker.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    expect(minutes59).toBeInTheDocument();
+  });
+  it("allows to change minutes format for ten minutes", async () => {
+    const view = new TimePickerObject().render({
+      minutesFormat: "tenMinutes",
+    });
+
+    view.focusInput();
+    const minutes50 = view.timePart("minutes", "50");
```
monikaracka, line 116: What about `expectedMinutes`?

packages/design-system/src/components/TimePicker/__tests__/TimePicker.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+    expect(minutes50).toBeInTheDocument();
+  });
+  it("display seconds when timeFormat is hh:mm:ss", async () => {
+    const view = new TimePickerObject().render({
+      secondsFormat: "seconds",
+      timeFormat: "hh:mm:ss",
+    });
+
+    view.focusInput();
+    const seconds59 = view.timePart("seconds", "59");
```
monikaracka, line 127: What about `expectedSeconds`?

packages/design-system/src/components/TimePicker/useTimePickerHelpers.ts
```diff
@@ -4,51 +4,45 @@ import {
...
+        values.push(i.toString());
+      }
+    }
+  }
+  return values;
+};
+
+const every1 = getSecondsOrMinutes(1);
+const every10 = getSecondsOrMinutes(10);
+
```
monikaracka, line 23: The earlier solution for me looked clearer and faster because we didn't have to call a function to return a short array. We don't need to have a function for every interval. 
Apart from that, I don't particularly appreciate using numbers in const names. I found no information that we shouldn't do this, but I think writing the number in letters would be nicer (e.g., everyOneMinuteArray).
And I see you didn't use const for quarters, and it would be the best solution to avoid these variable names.

packages/design-system/src/components/TimePicker/__tests__/TimePicker.screen-object.tsx
```diff
@@ -0,0 +1,116 @@
...
+    return screen.queryByRole("button", { name: "PM" });
+  }
+  get OKButton() {
+    return screen.getByRole("button", { name: "OK" });
+  }
+  get cancelButton() {
+    return screen.getByRole("button", { name: "Cancel" });
+  }
+
+  scroll(timePart: TimePickerTimeParts): HTMLUListElement {
```
monikaracka, line 66:
```suggestion
  getScroll(timePart: TimePickerTimeParts): HTMLUListElement {
```
This is a function, so we should use verbs as prefixes to imply that an action is taking place.

packages/design-system/src/components/TimePicker/__tests__/TimePicker.screen-object.tsx
```diff
@@ -0,0 +1,116 @@
...
+    return screen.getByTestId(`${timePart}Scroll`);
+  }
+  timePart(timePart: TimePickerTimeParts, timeText: string) {
```
monikaracka, line 69: here the same

packages/design-system/src/components/TimePicker/__tests__/TimePicker.screen-object.tsx
```diff
@@ -0,0 +1,116 @@
...
+    //eslint disabled cause it says "Avoid destructuring queries from `render` result, use `screen.getByText` instead"
+    //we wan't to search for it in secific scroll as values may be repeated in different scrolls
+    // eslint-disable-next-line testing-library/prefer-screen-queries
+    return getByText(this.scroll(timePart), timeText);
+  }
+
+  queryTimePart(timePart: TimePickerTimeParts, timeText: string) {
```
monikaracka, line 76: here the same

zuzannanap: query is a verb, so i guess in this case it's ok ;)

monikaracka (approved):

PR #452, opened 2023-08-08
Ds 695 social buttons unit tests

zuzannanap:


packages/design-system/src/stories/examples/SocialButtons/AppleButton.stories.tsx
```diff
@@ -25,6 +25,7 @@ export default {
     href: "https://www.Apple.com",
     variant: "black",
     disabled: false,
+    onClick: console.log("button clicked"),
```
monikaracka, line 28: I think it's not necessary here. You can test if onClick works.

PR #467, opened 2023-08-11
fix/DS-718 - sidebar

monikaracka:


packages/design-system/src/components/Sidebar/ResponsiveSidebar.tsx
```diff
@@ -22,34 +23,34 @@ export const ResponsiveSidebar: FC<ResponsiveSidebarProps> = (props) => {
     ...rest
   } = props;
 
-  const [isOpen, setIsOpen] = useState(
-    rest.isOpen !== undefined ? rest.isOpen : defaultIsOpen
+  const [isSideBarOpen, setIsSideBarOpen] = useState(
```
zuzannanap, line 26: Sidebar instead of SideBar

zuzannanap (approved):

PR #470, opened 2023-08-14
change borders and spacings of AvatarGroup

zuzannanap:


packages/design-system/src/components/AvatarGroup/index.tsx
```diff
@@ -13,14 +13,26 @@ export const AvatarGroup: React.FC<AvatarGroupProps> = ({
...
 
   return (
-    <div className={classNames("flex -space-x-3", className)}>
+    <div
+      className={classNames(
+        "flex",
+        className,
+        (size === "xxs" || size === "xs") && "-space-x-2",
+        (size === "sm" || size === "md") && "-space-x-3.5",
+        (size === "lg" || size === "xl" || size === "xxl") && "-space-x-3"
```
monikaracka, line 22: Create an object with variants to keep convention from other components.

packages/design-system/src/components/AvatarGroup/index.tsx
```diff
@@ -13,14 +13,26 @@ export const AvatarGroup: React.FC<AvatarGroupProps> = ({
...
       {visibleAvatarsArray.map((avatar, index) => (
         <Avatar
           {...avatar}
           key={`${index}-id`}
           size={size}
           shape={shape}
           contentType={contentType}
+          className={classNames(
+            "border-avatarGroup-base",
+            size === "xxs" || size === "xs" ? "border-[1.5px]" : "border-2"
```
monikaracka, line 34: What about create `const extraSmallVariants = size === "xxs" || size === "xs"`
```suggestion
            extraSmallVariants ? "border-[1.5px]" : "border-2"
```

or `const borderWidthVariants = size === "xxs" || size === "xs" ? "border-[1.5px]" : "border-2"`
```suggestion
           borderWidthVariants
```

packages/design-system/src/components/AvatarGroup/index.tsx
```diff
@@ -30,6 +42,10 @@ export const AvatarGroup: React.FC<AvatarGroupProps> = ({
           shape={shape}
           size={size}
           contentType="text"
+          className={classNames(
+            "border-avatarGroup-base",
+            size === "xxs" || size === "xs" ? "border-[1.5px]" : "border-2"
```
monikaracka, line 47: Here you could use the same.

monikaracka (approved):

PR #472, opened 2023-08-14
DS 717 - FileUploads error messages

zuzannanap:


monikaracka (commented):
Screenshot 2023-08-16 at 10 32 27
I'm not sure we should change supporting text `SVG, PNG, JPG or GIF. Max size of 2MB ` to an error message. Probably we have to wait for Paweł to get the answer.

packages/design-system/src/components/FileUpload/SmallFileUpload.tsx
```diff
@@ -67,15 +53,15 @@ export const SmallFileUpload: React.FC<SmallFileUploadProps> = ({
...
-        </p>
+        <InputText
+          text={errorMessage || supportingText}
+          type={getInputBottomTextType({
+            isError,
+            isDisabled: disabled,
+          })}
+          id={id}
+          size="xs"
+        />
```
jakmarch, line 64: According to the designs, supporting text/err msg should be `text-sm`

jakmarch (changes requested):

pauzim:
What happens with this PR? 🤔

PR #474, opened 2023-08-14
DS726: fix disabled state in Toggle

zuzannanap:


packages/design-system/src/components/Toggle/index.tsx
```diff
@@ -49,7 +49,7 @@ export const Toggle = forwardRef<HTMLInputElement, ToggleProps>(
           />
           <button
             aria-pressed={checked}
-            onClick={() => onChange(checked)}
+            onClick={() => !disabled && onChange(checked)}
```
monikaracka, line 52: You can pass `disabled={disabled}` to `button` instead of this solution. 

Could you change `peer-disabled:cursor-not-allowed` to `peer-disabled:cursor-auto`?

monikaracka (approved):

PR #478, opened 2023-08-14
Ds 725 toasts fixes

zuzannanap:


packages/design-system/src/components/Toast/index.tsx
```diff
@@ -96,7 +96,7 @@ export const ToastComponents = {
     return (
       <p
         className={classNames(
-          "empty:hidden",
+          "break-words empty:hidden",
```
monikaracka, line 99: I checked and I have this style, but it doesn't work for me. `break-all` works. 
Nevertheless, I wonder if breaking words is a good idea in this case. It shouldn't be a very long message in a toast and I think it doesn't not so many very long words. I'm afraid that some words that we don't want to break could be broken.

monikaracka (approved):

PR #483, opened 2023-08-16
DS-723: Radio aria-describedby 

zuzannanap:


packages/design-system/src/components/Radio/index.tsx
```diff
@@ -64,7 +63,7 @@ export const Radio = forwardRef<HTMLInputElement, RadioProps>(
           checked={checked}
           value={value}
           onChange={onChange}
-          aria-describedby={ariaDescribedBy}
+          aria-describedby={!!supportingText ? `${id}-supportingText` : ""}
```
monikaracka, line 66: You don't have to use `!!` before props. If you don't use it, the result will be the same.

monikaracka (approved):

PR #491, opened 2023-08-17
DS-742: TimePicker AM/PM

zuzannanap:


monikaracka (commented):
Timepicker could also have a dropdown inside input ( `inputVariant = "withDropdown"` ) and now it doesn't work well.

monikaracka (approved):

PR #496, opened 2023-08-18
Ds-745-Select Fixes

zuzannanap:


packages/design-system/src/components/Dropdown/SelectDropdownComponents.tsx
```diff
@@ -213,6 +212,7 @@ export const SelectDropdownComponents = {
     );
   },
   Input: ({ id, ...props }: InputProps<IDropdownOption>) => {
+    //@ts-ignore
```
monikaracka, line 215: Why did you add this here?  I don't see here any problem whit types.

monikaracka (approved):

packages/design-system/src/components/Dropdown/SelectDropdownComponents.tsx
zuzannanap: it started to throw error in my console for some reason

PR #500, opened 2023-08-18
DS-637: enable RangeSlider width change

zuzannanap:


packages/design-system/src/components/Dropdown/SelectDropdownComponents.tsx
```diff
@@ -213,6 +213,7 @@ export const SelectDropdownComponents = {
     );
   },
   Input: ({ id, ...props }: InputProps<IDropdownOption>) => {
+    //@ts-ignore
```
monikaracka, line 216: Is this necessary?

monikaracka (commented):
What about w-full? Then in className, we can pass max-w-[some value]? In this way, we could avoid important in class.

packages/design-system/src/components/Dropdown/SelectDropdownComponents.tsx
zuzannanap: yeeeessss :d

zuzannanap:
> 

good idea :)

packages/design-system/src/stories/utils.tsx
```diff
@@ -319,3 +319,22 @@ export const componentHeadingControl = {
   options: Object.keys(componentHeadingList),
   mapping: componentHeadingList,
 };
+
+export const rangeSliderClassNameOptions = {
+  width200: "!w-[200px]",
```
monikaracka, line 324: You don't have to use width class with `important`. Use `max-width` class. It should work

monikaracka (approved):

PR #502, opened 2023-08-18
DS-712 - unify components

monikaracka:
It's not good idea to merge it to main because a lot of props names are different so it could be in second version of TACO.


packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
```diff
@@ -79,15 +78,16 @@ export const AsyncBaseDropdown = forwardRef<
...
           loadingMessage={() => loadingMessage}
           inputId={id}
           {...props}
         />
         <div className={classNames("empty:hidden", isError && "min-h-6")}>
           <InputText
             text={errorMessage || supportingText}
-            type={getInputBottomTextType({ isError, isDisabled })}
+            error={isError}
+            disabled={disabled}
```
zuzannanap, line 90: just 'disabled' will be enough

packages/design-system/src/components/DatePicker/DatePickerComponents.tsx
```diff
@@ -124,18 +128,18 @@ export const DatePickerComponents = {
...
         />
         <InputText
-          type={getInputBottomTextType({
-            isError,
-            isDisabled: disabled,
-          })}
           id={`${inputId}-inputText`}
           text={errorMessage || supportingText}
+          error={isError}
+          disabled={disabled}
```
zuzannanap, line 142: just 'disabled' will be enough

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
```diff
@@ -48,15 +47,15 @@ export const AsyncBaseDropdown = forwardRef<
     const [isSelected, setIsSelected] = useState<boolean>(false);
     const { components } = useRenderCustomComponents({
       isClearable,
-      isDisabled,
+      isDisabled: disabled,
```
zuzannanap, line 50: when i see it i think that isDisabled would be better than disabled - we use isSomething for booleans everywhere

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
```diff
@@ -79,15 +78,16 @@ export const AsyncBaseDropdown = forwardRef<
...
+          isDisabled={disabled}
           loadingMessage={() => loadingMessage}
           inputId={id}
           {...props}
         />
         <div className={classNames("empty:hidden", isError && "min-h-6")}>
           <InputText
             text={errorMessage || supportingText}
-            type={getInputBottomTextType({ isError, isDisabled })}
+            error={isError}
```
zuzannanap, line 89: an here isError would be better i think

packages/design-system/src/components/Dropdown/SelectDropdown/SelectDropdown.tsx
```diff
@@ -79,7 +78,8 @@ export const BaseDropdown = forwardRef<HTMLSelectElement, ISelectDropdownProps>(
         />
         <InputText
           text={errorMessage || supportingText}
-          type={getInputBottomTextType({ isError, isDisabled })}
+          error={isError}
+          disabled={disabled}
```
zuzannanap, line 82: same

packages/design-system/src/components/Input/index.tsx
```diff
@@ -62,7 +61,8 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
         </InputComponents.BaseInput>
         <InputText
           text={errorMessage || supportingText}
-          type={getInputBottomTextType({ isError, isDisabled: disabled })}
+          error={isError}
+          disabled={disabled}
```
zuzannanap, line 65: same

packages/design-system/src/components/InputAutocomplete/index.tsx
```diff
@@ -150,10 +149,8 @@ export const InputAutocomplete = forwardRef<
           <Dropdown inputValue={value} {...props} />
         </Input.BaseInput>
         <InputText
-          type={getInputBottomTextType({
-            isError,
-            isDisabled: disabled,
-          })}
+          error={isError}
+          disabled={disabled}
```
zuzannanap, line 153: same

packages/design-system/src/components/InputCode/index.tsx
```diff
@@ -200,7 +205,8 @@ const InputCode: FC<InputCodeProps> = ({
       </div>
       <InputText
         text={errorMessage || supportingText}
-        type={getInputBottomTextType({ isError, isDisabled: disabled })}
+        error={isError}
+        disabled={disabled}
```
zuzannanap, line 209: same

packages/design-system/src/components/InputNumber/index.tsx
```diff
@@ -144,7 +146,8 @@ const InputNumber = forwardRef<HTMLInputElement, InputNumberProps>(
 
         <InputText
           text={errorMessage || supportingText}
-          type={getInputBottomTextType({ isError, isDisabled: disabled })}
+          error={isError}
+          disabled={disabled}
```
zuzannanap, line 150: same

packages/design-system/src/components/InputQuantity/index.tsx
```diff
@@ -119,7 +118,8 @@ const InputQuantity = forwardRef<HTMLInputElement, InputQuantityProps>(
 
           <InputText
             text={errorMessage || supportingText}
-            type={getInputBottomTextType({ isError, isDisabled: disabled })}
+            error={isError}
+            disabled={disabled}
```
zuzannanap, line 122: same... :d

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
monikaracka: I disagree because we have to pass the value of `disabled`. If we pass only `disabled` instead of `disabled={disabled}`, it always will be true.

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
monikaracka: We usually use `disabled` or `error`  in props like in HTML elements props. `isSomething` we use when we check something is true, and here for arguments in function is better to use `isSomething`, but we have to pass `disabled` from props.

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
monikaracka: It's the same as with disabled in the first example.

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
zuzannanap: no, because it's not a prop of a component, it's an argument for a function here

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
zuzannanap: ok, i agree :)

packages/design-system/src/components/Dropdown/AsyncSelect/AsyncSelect.tsx
monikaracka: It's not function anymore.

zuzannanap: ok i'm an idiot🙃 to early for a review xd

zuzannanap (approved):

PR #503, opened 2023-08-18
DS-746 - input quantity step

monikaracka:


zuzannanap:
when step is not an integer this happens:
image

zuzannanap (approved):

PR #521, opened 2023-08-24
DS-765-RadioCardGroup jumping because of the border

zuzannanap:


packages/design-system/src/components/RadioCard/index.tsx
```diff
@@ -68,11 +68,7 @@ export const RadioCard = forwardRef<HTMLInputElement, RadioCardProps>(
           "flex min-w-fit cursor-pointer items-center rounded-radioCard-radiusBase border bg-radioCard-base p-radioCard-paddingLabel [&:has(:focus-visible)]:ring-4 [&:has(:focus-visible)]:ring-radioCard-base-focused",
           positionVariantStyles[radioPosition],
           alignVariantStyles[align],
-          {
-            "!border-2": disabled && checked,
-            "hover:shadow-radioCard": !disabled,
-          },
-
+          disabled && "hover:shadow-radioCard",
```
monikaracka, line 71: What does this style do?  Hover on disabled?

monikaracka (commented):
Screenshot 2023-08-25 at 08 34 06

And by the way it's something wrong with supporting text color when it's disabled.

packages/design-system/src/components/RadioCard/index.tsx
```diff
@@ -68,11 +68,8 @@ export const RadioCard = forwardRef<HTMLInputElement, RadioCardProps>(
           "flex min-w-fit cursor-pointer items-center rounded-radioCard-radiusBase border bg-radioCard-base p-radioCard-paddingLabel [&:has(:focus-visible)]:ring-4 [&:has(:focus-visible)]:ring-radioCard-base-focused",
           positionVariantStyles[radioPosition],
           alignVariantStyles[align],
-          {
-            "!border-2": disabled && checked,
-            "hover:shadow-radioCard": !disabled,
-          },
-
+          disabled && checked && "border-2",
```
monikaracka, line 71:
```suggestion
          checked && "border-2",
```

monikaracka (approved):

packages/design-system/src/components/RadioCard/index.tsx
zuzannanap: it's ok, i wanted it to be thick when disabled and checked at the same time

PR #522, opened 2023-08-24
Ds 767: Button bugs

zuzannanap:


monikaracka (approved):
I think it would be better to use Button in this way:
 `<Button>text</Button>` 
 instead of 
`<Button children="text"/>`
but the decision is on you.

PR #524, opened 2023-08-25
DS-766: Scrollable Modal

zuzannanap:


packages/design-system/src/stories/examples/ScrollableModal.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <button onClick={() => setOpen(true)}>Open Modal</button>
+      <ModalComponent
+        {...props}
+        isOpen={open}
+        onClose={() => setOpen(false)}
+        className="max-h-screen w-6/12"
```
monikaracka, line 21: Maybe `w-1/2`? It would look better for me. But the decision is on you.

monikaracka (approved):

PR #525, opened 2023-08-25
DS 754: dont hide input from the DOM

zuzannanap:


packages/design-system/src/components/RadioCard/__tests__/RadioCard.screen-object.tsx
```diff
@@ -29,6 +29,9 @@ export class RadioCardObject {
   get radioVisibleElement() {
     return screen.queryByTestId("radio");
   }
```
monikaracka, line 31: I think it's not used here.

monikaracka (approved):

PR #530, opened 2023-08-28
DS-737 - date picker icon

monikaracka:


packages/design-system/src/components/DatePicker/DatePicker.tsx
```diff
@@ -36,6 +36,7 @@ const DatePicker: FC<DatePickerProps> = ({
   disabled,
   ...props
 }) => {
+  const [open, setOpen] = useState<boolean>(false);
```
zuzannanap, line 39: isOpen/setIsOpen ?

zuzannanap (approved):

PR #540, opened 2023-09-07
DS-752 - Select dropdown types fix

jakmarch:
Now, onSelect() value depends on isMulti prop value.


zuzannanap (approved):

pauzim:
Can we merge it or we are waiting for sth? 🤔

PR #541, opened 2023-09-07
Ds 610: aria attributes for components with icon button

zuzannanap:


packages/design-system/src/components/VideoPlayer/VideoPlayer.interface.ts
```diff
@@ -1,18 +1,11 @@
...
-  icon: TComponentIcon;
-  onClick(event: MouseEvent<HTMLButtonElement>): void;
-};
-
-export interface IControlsButtonProps {
-  icon: TComponentIcon;
-  onClick(event: MouseEvent<HTMLButtonElement>): void;
-}
+export interface IControlsButtonProps
+  extends Required<Pick<IconButtonProps, "aria-label" | "icon" | "onClick">> {}
```
pauzim, line 5: Let's remove `Required` -> with this there's a strange situation because:
- `controls` prop is not required;
- but if I want to pass any of controls props, then I'm obligated to pass all of them because they are required

pauzim (approved):

PR #542, opened 2023-09-08
Ds 610 aria attributes toggle

zuzannanap:


packages/design-system/src/components/Toggle/Toggle.interface.ts
```diff
@@ -2,15 +2,18 @@ import { ClassName } from "@helpers/types";
 import { InputHTMLAttributes } from "react";
 
 export interface ToggleProps
-  extends Omit<InputHTMLAttributes<HTMLInputElement>, "size" | "onChange">,
+  extends Omit<
+      InputHTMLAttributes<HTMLInputElement>,
+      "size" | "onChange" | "aria-label"
+    >,
+    Required<Pick<InputHTMLAttributes<HTMLInputElement>, "aria-label">>,
```
pauzim, line 9: Let's not require  `aria-label`

pauzim (approved):

PR #543, opened 2023-09-08
Ds 610 aria attributes part 3

zuzannanap:


packages/design-system/src/components/TimePicker/TimePicker.interface.ts
```diff
@@ -66,12 +72,14 @@ export interface TimePickerDropdownProps extends ClassName {
...
-  secondsStartScrollIndex?: number;
+  hoursScrollProps?: ScrollProps;
+  minutesScrollProps?: ScrollProps;
+  secondsScrollProps?: ScrollProps;
 }
 
-export interface ScrollListProps extends PropsWithChildren {
+export interface ScrollListProps
+  extends PropsWithChildren,
+    Required<Pick<HTMLAttributes<HTMLUListElement>, "aria-label">> {
```
pauzim, line 82: Should it be required?

PR #547, opened 2023-09-11
DS 786: enable ReactNode instead of <a> tag

zuzannanap:


packages/design-system/src/stories/molecules/FeaturedText.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import { FeaturedTextProps } from "@components/FeaturedText/Featuredtext.interface";
 import FeaturedTextComponent from "@components/FeaturedText";
 import { storybookIconImgControl } from "../utils";
+import { ReactNode } from "react";
```
piotrekzawada, line 4: unused import?

packages/design-system/src/components/Anchor/index.tsx
```diff
@@ -16,17 +18,15 @@ export const Anchor: React.FC<AnchorProps> = ({
...
         "text-sm",
         decorationVariants[decoration],
         className
       )}
-      href={link}
-      role={role}
     >
-      {text}
-    </a>
+      <LinkTag href={href}>{children}</LinkTag>
```
piotrekzawada, line 28: shouldn't we simply do sth like this:
```suggestion
      {children}
```

PR #548, opened 2023-09-13
Ds 793 input autocomplete fixes

zuzannanap:


packages/design-system/src/components/InputAutocomplete/InputAutocomplete.interface.ts
```diff
@@ -1,26 +1,35 @@
...
 export interface InputAutocompleteProps
-  extends Omit<DropdownProps, "inputValue">,
+  extends Omit<
+      DropdownProps,
+      "inputValue" | "onItemClick" | "setIsDropdownVisible"
+    >,
     Omit<InputProps, "value"> {
-  value: string;
   onClear?(): void;
+  setValue: Dispatch<SetStateAction<string>>;
```
kielbasa-elp, line 33: lets change the name to onSelect.
I am not sure if typing this function by using `Dispatch<...> `is a good idea. Maybe the better option is to declare the type `(value: string) =>void` ?

packages/design-system/src/components/InputAutocomplete/Dropdown.tsx
```diff
@@ -32,41 +34,65 @@ export const DropdownItem: FC<DropdownListItemProps> = ({
...
           <DropdownItem
-            label={label}
+            value={value}
             icon={icon}
-            value={inputValue}
-            key={`${label}-${index}`}
+            inputValue={inputValue}
+            key={`${value}-${index}`}
+            onClick={onItemClick}
+            setIsDropdownVisible={setIsDropdownVisible}
```
kielbasa-elp, line 95: Instead of passing setter, just pass the onClose func 
eg. `function onClose(){ setIsDropdownVisible(false)}`

kielbasa-elp (approved):

PR #558, opened 2023-09-21
Ds 769 dropdowns portals

zuzannanap:


pauzim (approved):
tbh it's hard to test it inside storybook since it has some nested document but it looks like everything's OK. Mounts correctly to the nearest document.body and behaves properly too (overflows even with `overflow:hidden` set) 👍

pauzim:
Should we add this also to `SelectDropdown`, `Async` and `Creatable` ? If so, let's do this in a separate PR 🙏

PR #564, opened 2023-09-25
DS 781: fix icon position in iconButton

zuzannanap:


pauzim (commented):
I think we should keep icon as `span` element- if necessary with some `inline-block` class. If we switch to `div` , icon will be by default displayed under eg. additional text instead of one line which is the most common case
image

pauzim (approved):

PR #570, opened 2023-09-27
Resolve conflicts

pauzim:


packages/design-system/src/stories/molecules/TimePicker/TimePicker.stories.tsx
```diff
@@ -12,9 +13,16 @@ import {
...
-    defaultTimePickerDropdownTime
-  );
+  const defaultDropdownTime = {
+    hours: 1,
+    minutes: "00",
+    seconds: "00",
+    period: PeriodTypes.AM,
+  };
+
+  const inputRef = useRef<HTMLInputElement>(null);
```
zuzannanap, line 23: now we have ref from forwardref instead of this inputRef

PR #574, opened 2023-09-28
Ds 749 refactor

zuzannanap:


packages/design-system/src/components/Pagination/PaginationComponents.tsx
```diff
@@ -0,0 +1,118 @@
...
+          variant === "default" &&
+          "-ml-pagination-marginButton rounded-r-pagination-radiusBase"
+      )}
+      onClick={onClick}
+      disabled={disabled}
+    >
+      {direction === "left" ? (
+        <RightArrowSvg className="aspect-square w-5 -rotate-180 transform" />
+      ) : (
+        <RightArrowSvg className="order-last aspect-square w-5" />
```
pauzim, line 73: we can render `RightArrowSvg` just once and handle `rotate transform` claases based on `direction` prop

pauzim (approved):

packages/design-system/src/components/Pagination/PaginationComponents.tsx
zuzannanap: i made it in another pr before :)

PR #575, opened 2023-09-28
Rename: `DropdownMenu` -> `DropdownMenu.Container`

pauzim:
Change naming in storybook, because currently it's somehow misleading - based on feedback - people think it's some separate component.


zuzannanap (approved):
It's ok, but do we need it in our storybook at all?..🤔

pauzim:
> It's ok, but do we need it in our storybook at all?..🤔

I think we could keep this as one of available variants of implementing dropdowns with design system. Actually I think this example in the story is the lightest one 😅

PR #579, opened 2023-10-01
DS 812: refactor Dropdowns files structure

zuzannanap:


packages/design-system/src/components/Dropdown/MenuDropdown/MenuDropdownComponents.tsx
```diff
@@ -8,8 +8,8 @@ import {
   IUserData,
   IDropdownProps,
 } from "./MenuDropdown.interface";
-import { DropdownComponents } from "../DropdownComponents";
-import { IDropdownOption } from "../Dropdown.interface";
+import { SelectDropdownComponents } from "../Select/SelectDropdownComponents";
```
pauzim, line 11: I think I'd like to keep `DropdownComponents` as they were previously - the purpose was to have some file with shared components between both `Select...` and `Menu...` dropdowns. If you feel the naming `Dropdown...` is misleading we could change it to eg sth like `SharedDropdownComponents`

PR #583, opened 2023-10-02
DS-809: Add `DatePicker` registered field, display validation errors

pauzim:


pauzim:
> Everything here is ok ✅ I know you didn't change this, but I'm wondering about the place of errors and submitted data. I couldn't find them right away. I thought that they'd appear under the submit button.

You think it'd be more intuitive to display them under submit button? I can change this

monikaracka:
> You think it'd be more intuitive to display them under submit button? I can change this

I expected this under the button, so yes, I think It would be more intuitive 😃

PR #599, opened 2023-10-04
DS-840: `customStyles` -> `customStylesOverrides` 

pauzim:
Not sure how about exporting some types from `react-select`, is this a good approach 🤔


packages/design-system/src/components/Dropdown/index.ts
```diff
@@ -55,4 +63,9 @@ export type {
   IDropdownOption,
   IOptionProps,
   ICreatableSelectProps,
+  StylesConfig,
+  GroupBase,
+  CSSObjectWithLabel,
+  OptionProps,
+  PortalStyleArgs,
```
monikaracka, line 70: I wonder... If we want to use `Selects` in DS, we need to install the `react-select` package. So, we could import types directly from React-select. If this is true, we won't need to export react-select types.

zuzannanap (approved):

monikaracka (approved):

kielbasa-elp (approved):

PR #617, opened 2023-10-10
DS-823 - radio card group variants

monikaracka:


pauzim (commented):
- Are we using a flag `isRadioVisible` for now? I can't find it in component's code, but it's present in interface `RadioCardGroupProps` and in storybook (`RadioCardGroup.stories`) and it looks like it doesn't change anything 🤔
- We have to pass `children` prop (as children) to `RadioCard` component

packages/design-system/src/components/RadioCardGroup/RadioCardGroup.interface.ts
```diff
@@ -3,13 +3,15 @@ import {
...
 
 export interface RadioCardGroupOption {
   id: string;
+  value: string;
   checked?: boolean;
-  labelText: string;
+  labelText?: string;
   description?: string;
-  value: string;
+  children?: ReactNode;
```
zuzannanap, line 14: maybe "extends PropsWithChildren" instead?

zuzannanap: and actually it might just extend a part of interface of RadioCardProps

monikaracka: But this is the interface of an object, not props, so this could not be very clear.

pauzim (approved):

PR #621, opened 2023-10-10
DS-827: `SelectDropdown` stories

pauzim:
Just a docs
No breaking changes


monikaracka (approved):
Great job! 🚀 
We have one thing to consider after this PR. Is it better to show docs after or before the table with props 🤔 ?

pauzim:
> Great job! 🚀 We have one thing to consider after this PR. Is it better to show docs after or before the table with props 🤔 ?

Agree! let's decide in next PR I had the same thoughts

PR #623, opened 2023-10-10
DS-852: fix input search - wrong styles after setting addonsVariant prop

olborska-elp:


packages/design-system/src/components/Search/index.tsx
```diff
@@ -48,6 +48,14 @@ export const Search = forwardRef<HTMLInputElement, SearchProps>(
 
     const isInline = variant === "inline";
 
+    const isInputPrefixShown =
```
zuzannanap, line 51: why not  props.addonsVariant === InputAddonsVariants.PREFIX || props.addonsVariant === InputAddonsVariants.PREFIX_AND_SUFFIX ?
I think it would be a bit clearer

zuzannanap (approved):

packages/design-system/src/components/Search/index.tsx
```diff
@@ -86,35 +94,42 @@ export const Search = forwardRef<HTMLInputElement, SearchProps>(
...
-                <CircularCloseSvg className="h-4 w-4" />
-              )
-            }
-            onClick={isIconVisible ? handleClear : undefined}
-            className={addonsStyles}
-            error={error}
-            aria-label="clear"
-            {...suffixProps}
-          />
+          {isInputPrefixShown && (
```
zuzannanap, line 108: Prefix is always displayed in search

olborska-elp: It's a good idea for me to show always prefix for search but in this case we need to change interface - now `addonsVariant?: InputAddonsVariants` (`InputAddonsVariants` has 4 options).
I was wondering too if we can pass a prop with an icon for prefix (now it is hardcoded with magnifying glass icon).
Maybe we could discuss both cases and make changes in the next PR.

packages/design-system/src/components/Search/index.tsx
olborska-elp: If we will use `strict equality` operator instead of `strict inequality`, we should add third condition - `props.addonsVariant === undefined` (`props.addonsVariant` is by default `undefined` if we don't set props), that's why I used `inequality` - so that there are fewer conditions

packages/design-system/src/components/Search/index.tsx
pauzim: Let's merge it how it is for now since it fixs main issue and go back to `icon` discussion in next task 🤔

pauzim (approved):

PR #627, opened 2023-10-10
DS 722 Pagination refactor

zuzannanap:


olborska-elp (approved):
It looks good for me and I like refactor changes - code is more clearer and readable now.

Examples in Storybook works consistently - but it would be good if it was checked by someone with more experience in design system (design & logic).
- The only tricky part for me is how works `rangeSize` prop - how it changes number of page buttons in various cases (we can catch up for huddle if you would like to discuss this case)

What do you think about changing styles of arrow buttons if they're disabled? Now we can't click them but sometimes there's a convention, that these buttons are lighter/darker/ other styles (I understand that it may be out of scope of this task).

Screenshot 2023-10-11 at 12 15 04

pauzim:
Please wait with merge until next publish (post 0.5.0 taco version -> let's publish only aria labels changes in `0.5.0`)

PR #634, opened 2023-10-11
DS 860: Modal docs

zuzannanap:


packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -18,15 +18,15 @@ const Modal: FC<ModalProps> = ({
   imgSrc,
   ...rest
 }) => {
-  const isHeader =
-    header || closeButtonProps?.icon || closeButtonProps?.iconName;
+  const isHeader = header || !!closeButtonProps;
+  const isLargeVariant = size === "lg";
 
-  const variantsStyle = {
+  const styleVariants = {
```
monikaracka, line 24: The previous name was more appropriate. These are styles for different variants.

packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -18,15 +18,15 @@ const Modal: FC<ModalProps> = ({
...
     shadow: "shadow-modal-shadowBase",
     outline: "border-modal-borderWidthBase border-modal-outline",
   };
 
-  const sizeVariantsStyle = {
+  const sizeVariants = {
```
monikaracka, line 29: These aren't size variants but styles in terms of size variants.

packages/design-system/src/stories/atoms/Modal/Modal.docs.tsx
```diff
@@ -0,0 +1,82 @@
+import { StoryDocsCodeBlock } from "stories/docsHelpers/StoryDocsCodeBlock";
+
+export const ModalDocs = () => (
+  <div>
+    <h1>How to use DS Pagination</h1>
+    <p>
+      To use TimePicker, you have to pass <b>required props</b> to the
+      component:
```
monikaracka, line 8: Pagination? Timepicker? 😉

zuzannanap: ok, double fail :D

monikaracka (approved):

PR #636, opened 2023-10-11
DS-819 - `sidebar` changes

monikaracka:


pauzim (approved):
One thing imo is missing - update `Sidebar` stories with some collapse state management- sth like this:

```
const [isCollapsed, setIsCollapsed] = useState<boolean>();

```
Also, if we want to storybook's control `collapsed` to work, we can add some useEffect to cause rerenders:

```
  useEffect(() => {
    setIsCollapsed(props.collapsed);
  }, [props.collapsed]);
```

PR #637, opened 2023-10-12
DS-859: search with custom icon

olborska-elp:
- add custom icon to input prefix
- add custom icon to icon button inside search
- add custom text to button inside search

I hope - no breaking changes.

Feel free to CR notes about names or used conventions in taco.


packages/design-system/src/components/Search/Search.interface.ts
```diff
@@ -19,4 +27,7 @@ export interface SearchProps
   suffixProps?: InputAddonProps;
   buttonProps?: SearchButtonProps;
   iconButtonProps?: SearchIconButtonProps;
+  buttonText?: string;
+  prefixIcon?: TSearchIcon;
+  iconForIconButton?: TSearchIcon;
```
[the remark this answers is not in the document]
pauzim: In components where we enable passing icon by its name or the icon component directly, we usually split props to `icon` and `iconName` (take a look at `IconButtonProps`). Not really sure which approach is better but maybe we could keep consistency in here and extend `IconButtonProps` in `iconButtonProps` object's interface ?

packages/design-system/src/components/Search/index.tsx
```diff
@@ -78,6 +94,16 @@ export const Search = forwardRef<HTMLInputElement, SearchProps>(
       }
     };
 
+    const defineIcon = useCallback(
+      (iconValue: TSearchIcon | undefined) => {
+        if (!iconValue)
+          return <SearchSvg className="block aspect-square h-4" />;
+        if (isValidElement(iconValue)) return iconValue;
+        return <Icon iconName={iconValue} />;
```
pauzim, line 102: my linter complains about `iconName` type  😓

packages/design-system/src/stories/utils.tsx
```diff
@@ -176,6 +181,25 @@ export const storybookChipsNodeElementControl = {
   },
 };
 
+export const storybookSearchIconControl = {
```
pauzim, line 184: if it's only for search component, you can move this control to a separate file next to `Search.stories` or use `storybookIconControl` instead (if it fits well to your requirements).

packages/design-system/src/components/Search/index.tsx
```diff
@@ -70,6 +82,10 @@ export const Search = forwardRef<HTMLInputElement, SearchProps>(
       isIconVisible && "peer-focus:!text-search-base-prefixFocus"
     );
 
+    const definedButtonText = isNonEmptyString(buttonText)
+      ? buttonText
+      : "Search";
```
zuzannanap, line 87: maybe just buttonText || "Search" ?

olborska-elp: I tried to handle scenario when `buttonText` may be a string with many spaces: `"      "` - `isNonEmptyString` fn trims string before checking. For `buttonText || "Search"`  if `buttonText = "   "` it return `"   "`

pauzim (approved):

PR #641, opened 2023-10-12
fix/DS-870 - fixes after merge branch with tokens

monikaracka:


packages/design-system/src/helpers/tokens/rating.json
```diff
@@ -198,6 +198,26 @@
...
+  },
+  "heightMD": {
+    "height": {
+      "height": {
+        "default": {
+          "value": "{10}",
+          "type": "height"
+        }
+      }
+    }
```
zuzannanap, line 220: sth is still wron with height:
image

zuzannanap (changes requested):

packages/design-system/src/helpers/tokens/rating.json
monikaracka: Maybe try running your localhost again because `rating` looks good for me.

olborska-elp:
On my laptop pink ring looks quite good:

Screenshot 2023-10-16 at 07 19 53

PR #642, opened 2023-10-12
Ds 862 table docs + refactor

zuzannanap:
### Does this PR includes some breaking changes (e.g changes in interfaces, required props etc.)?: 

Some props changed names:
data -> items
tableFooterCell -> footer (it's outside the table ant it's not a cell :))
pagination -> customPagination
selectionColumnClassName -> rowSelectionColumnClassName
onSelectRows -> onRowsSelect (to be consistent with e.d. onPrevButtonClick etc.)
paginationPrevButtonTitle -> paginationPrevButtonText (to be consistent with pagination props)
paginationNextButtonTitle -> paginationNextButtonText (to be consistent with pagination props)

## Description:

I added 2 version of docs with and without built in pagination (the difference is only in code example)
While writing docs i changed some props which had non-obvious names


packages/design-system/src/stories/molecules/Table/TableWitBuiltInPagination.docs.tsx
```diff
@@ -0,0 +1,151 @@
...
+  <div>
+    <h1>How to use DS Table</h1>
+    <p>
+      To use Table, you have to pass <b>required props</b> to the component:
+    </p>
+    <ul className="pb-4">
+      <li>
+        <b className="!text-blue-400">columns</b> ➤ an array containing the
+        columns that will appear in the table. Each column is an object with
+        required fields: <b>name</b>, <b>id</b> and not required fields:
```
pauzim, line 13: If you prefer, you can use `<HighlihtHelper>` small util component instead of `<b>`

pauzim (approved):
It's `DS-820` task number btw 😃

Good work with these docs! I wonder if there are some common docs parts we could share between all of the `Table` examples as eg `TableCommon.docs` and if so, maybe we could do the other docs for the specific variants.

PR #643, opened 2023-10-12
Ds 871 file uploads fixes

zuzannanap:

adds withBorder prop to DragAndDropFileUpload

## Description:

branch fixes colors bugs, adds option with border


pauzim (approved):
LGTM
 Only one thing I've noticed - maybe we could add `cursor-not-allowed` class for disabled DnD's ?

Also, I think we should add some separate task to include at least one of upload components to the `ExampleForm` to be sure it works well with `hook-form`

PR #647, opened 2023-10-13
Run rvt workflow on pull-request to the main branch

pauzim:
## Description:

Trigger visual regression tests (loki) by pull request to main.


pauzim:
To discuss - should we merge it or not - from one POV it seems to be more reasonable to run this workflow on pull request rather than merge to main because we can react before merge to the potential unexpected changes, on the other hand - for now we still have this issue with wrong snapshot creation related to the `Icon` compoents and some others and it looks like some infinite loop with triggering this workflow (because loki in most cases sees some diff and tries to produce the PR) 🤷‍♀️. And second - it looks like we still need this on push to `main` to protect unexpected direct pushes.

PR #654, opened 2023-10-13
DS-881: Input code rerender

pauzim:
## Description:

Little non-breaking logic update to rerender properly when `dashesPosition` prop changes (visible in storybook).


packages/design-system/src/stories/atoms/Inputs/InputCode/InputCode.stories.tsx
```diff
@@ -45,7 +46,7 @@ export default {
     label: "Label",
     supportingText: "supportingText",
     errorMessage: "",
-    dashesPositions: [1, 3],
+    dashesPositions: dashesInputCodeExampleOptions.multipleDashes,
```
olborska-elp, line 49: less magic numbers and magic arrays 😇

olborska-elp (approved):
Storybook example works without error 💥

PR #658, opened 2023-10-16
Ds 882 separate input text components

zuzannanap:
### Does this PR includes some breaking changes (e.g changes in interfaces, required props etc.)?: 

It adds Label and InputText classnames related to positioning only when prop hasInputFieldClassNames is set to true. 
this makes it possible to use Label and InputText components independently next to inputs, if we don't include label/supportingText/errorMessage props in the inputs

## Description:


packages/design-system/src/components/DatePicker/DatePickerComponents.tsx
```diff
@@ -142,7 +142,12 @@ export const DatePickerComponents = {
 
     return (
       <div className={className}>
-        <Label text={label} labelFor={inputId} disabled={disabled} />
+        <Label
+          text={label}
+          labelFor={inputId}
+          disabled={disabled}
+          belongsToCompleteInputField={true}
```
pauzim, line 149: OK, so in general -  I'm considering one thing, could we - instead of - `belongsToCompleteInputField` prop change a little bit `Label`'s component implementation and give className `empty:hidden` ? 🤔 In the optimistic scenario, if you don't provide any label `text`, the whole component should be hidden.

packages/design-system/src/components/InputText/index.tsx
```diff
@@ -29,7 +29,9 @@ export const InputText = ({
   return (
     <p
       className={classNames(
-        "text-inputText-fontBottomText col-start-1 col-end-3 m-0 mt-inputText-marginTop min-h-[6px] p-0 empty:hidden",
+        "text-inputText-fontBottomText empty:hidden",
+        belongsToCompleteInputField &&
+          "col-start-1 col-end-3 m-0 mt-inputText-marginTop min-h-[6px] p-0",
```
pauzim: Could be separated to some class

packages/design-system/src/components/Label/index.tsx
```diff
@@ -1,13 +1,21 @@
...
+}: LabelProps) => {
   if (!text) return null;
 
   const labelClassNames = classNames(
-    "block order-first col-start-1 col-end-3 m-0 mb-label-marginBottom min-h-[6px] p-0 empty:hidden",
+    "block empty:hidden",
     "text-sm font-medium",
     disabled ? "text-label-base-disabled" : "text-label-base",
+    belongsToCompleteInputField &&
+      "m-0 p-0 mb-label-marginBottom order-first col-start-1 col-end-3 min-h-[6px]",
```
pauzim: Same as above🤔

packages/design-system/src/components/Dropdown/Select/AsyncSelect/AsyncSelect.tsx
```diff
@@ -64,7 +64,11 @@ export const AsyncBaseDropdown = forwardRef<
 
     return (
       <div>
-        <Label text={label} disabled={disabled} />
+        <Label
+          text={label}
+          disabled={disabled}
+          belongsToCompleteInputField={true}
```
pauzim: You can shorten this in each place just `belongsToCompleteInputField`
```suggestion
          belongsToCompleteInputField
```

pauzim (approved):

PR #662, opened 2023-10-16
fix/DS-885 - `inputNumber`

monikaracka:


packages/design-system/src/stories/atoms/Inputs/InputNumber/InputNumber.stories.tsx
```diff
@@ -6,7 +6,9 @@ import { errorMessageControl } from "stories/utils";
 import { docs } from "./InputNumber.docs";
 
 export const InputNumber = ({ ...props }: InputNumberProps) => {
-  const [inputValue, setInputValue] = useState<number>(0);
+  const [inputValue, setInputValue] = useState<number>(
+    Number(props.defaultValue) || 0
```
zuzannanap, line 10: do we need this change if we can't change it with props anyway?

packages/design-system/src/stories/atoms/Inputs/InputNumber/InputNumber.stories.tsx
```diff
@@ -36,14 +38,17 @@ export default {
...
+    },
   },
   args: {
     label: "",
     supportingText: "",
-    defaultValue: 0,
     disabled: false,
     errorMessage: "",
     step: 1,
+    defaultValue: 0,
```
zuzannanap, line 51: do we need this line?

packages/design-system/src/stories/atoms/Inputs/InputNumber/InputNumber.stories.tsx
monikaracka: Yes, because I want to show how it can be used in code.

packages/design-system/src/stories/atoms/Inputs/InputNumber/InputNumber.stories.tsx
monikaracka: Yes, because I want to pass `defaultValue`.

packages/design-system/src/stories/atoms/Inputs/InputNumber/InputNumber.stories.tsx
pauzim: I didn't investigate the code so I might be wrong, but I think `InputNumber` renders `Input.BaseInput` which can take the `defaultValue` prop - maybe we could take a look why passing defaultValue doesn't work? 🤔 I remember we have some solution in the `SidePanel` component - that some passed default value (in that case defaultOpen or sth like this) is handled inside component 🤔

monikaracka: It works, and I think it works well. It should pass the value only at the beginning. After rendering, `defaultValue` shouldn't have any influence on the value.

pauzim: Ah Ok :D If we set initial state e.g to `0` and pass `defaultValue` to `3` -> the `3` will be displayed? If so, I'd suggest to remove `props.defaultValue` from `inputValue` state and pass it directly to the story component (I think then it would reflect the real usage?).

monikaracka: But if we don't pass it in that way, we can't take this value from state (`defaultValue` is equal to 4 here):
Screenshot 2023-10-17 at 09 51 53

zuzannanap (approved):

PR #682, opened 2023-10-20
fix/DS-894 - timepicker dropdown

monikaracka:


zuzannanap:
After closing dropdown for the first time without chaning time from 00:00, it changes to "1:00" in input

zuzannanap (approved):

PR #686, opened 2023-10-21
Ds 898 tokens docs

zuzannanap:

it changes exported tokens helpers (description in README :)) - now only helpers for "colors" and any "property" are exported

## Description:


packages/design-system/TOKENS.MD
```diff
@@ -0,0 +1,211 @@
...
+      }
+    }
+  }
+```
+
+This token sets the value for the class that sets the accordion's `fontWeight` property to `semibold`. To use this token in code, we use a class composed of `3 parts separated by dashes`:
+
+- tailwind prefix for a given property - here `font` (fontWeight in tailwind is set as e.g. `font-bold` or `font-normal`
+- the name of the `.json `file consistent with the name of the component - here it's `accordion`
+- token name from the `.json` file - here it't `title` (as we can see in the code fragment)
```
pauzim, line 31: The whole world was waiting for this instruction for sooooo long... 🥹🙏

packages/design-system/TOKENS.MD
```diff
@@ -0,0 +1,211 @@
...
+```
+
+### Add your own custom theme configures
+
+You have to merge the default theme and custom project config you created (Here we use for this purpose [lodash library merge method](https://lodash.com/docs/#merge) but you can do it any way you want.
+
+```js
+const { merge } = require('lodash');
+
+const customProjectConfig = {...}
```
pauzim, line 76: Is there some example of how to define this `customProjectConfig` object ? 🤔

packages/design-system/TOKENS.MD
```diff
@@ -0,0 +1,211 @@
...
+  ...
+};
+```
+
+It will add default DS token-based clases values to your project. They can bu customized to fit to your project design requirements.
+
+### How to use custom tokens values?
+
+- import appropriate tokenHelpers function
+- create your own config `.json` files with the configuration (in the future it will be automatically generated by `elp-figma-tokens-plugin`). Add it to your config with `getComponentsColorConfig()` token helper for colors and `getComponentsPropertyConfig(componentConfig, property, propertyType)` for all other properties (colors have one more level in `.json` config files, so this config is a bit different from others)
```
pauzim, line 90:
```suggestion
- create your own config `.json` files with the configuration (in the future it will be automatically generated by `elp-figma-tokens-plugin`). Add add it to your config with `getComponentsColorConfig()` token helper for colors and `getComponentsPropertyConfig(componentConfig, property, propertyType)` for all other properties (colors have one more level in `.json` config files, so this config is a bit different from others)
```

pauzim (approved):
Good job! 🙏

olborska-elp (approved):
It's great introduction to the tokens 🎉. Good job and I'm sure that this will be very helpful for future users and contributors.

I wonder if we can add somewhere (here or as extra file) an example of  entire `tailwind.config` file for the project (maybe based on HUB or other current project that uses Taco).
Snippets in `README` are very useful but sometimes it's tricky to put all parts together.

packages/design-system/TOKENS.MD
zuzannanap: yes, it's later in this manual - but true, it's not obvious - i'll added this information here :)

PR #691, opened 2023-10-23
Ds 903 scroll position in dropdown

zuzannanap:

it changes prop:
defaultDropdownTime -> defaultTime (boolean)
## Description:


packages/design-system/src/components/TimePicker/index.tsx
```diff
@@ -88,13 +93,19 @@ export const TimePicker = forwardRef<HTMLInputElement, TimePickerProps>(
...
+      isOKClick?: boolean
+    ) => {
       let value;
       if (isRefValid && ref.current) {
         const period = !withPeriodDropdown ? ` ${time.period}` : "";
-        const seconds = timeFormat === "hh:mm:ss" ? `:${time.seconds}` : "";
-        value = `${time.hours}:${time.minutes}${seconds}${period}`;
-        
+        const seconds =
+          timeFormat === "hh:mm:ss" ? `:${time.seconds || "00"}` : "";
```
monikaracka, line 104: It should be enum here.

packages/design-system/src/stories/examples/Form/ExampleForm.stories.tsx
```diff
@@ -180,7 +180,7 @@ export const ExampleForm = () => {
               aria-label=""
               supportingText="Choose time"
               id="timePicker"
-              defaultDropdownTime={defaultTimePickerDropdownTime}
+              defaultTime={true}
```
monikaracka, line 183: You don't have to pass `{true}`
```suggestion
              defaultTime
```

monikaracka (commented):
1. It doesn't work well. When I open the dropdown, the scroll positions are always `00` even `clearInputOnFocus` is false.
2.  We should refactor timepicker but that is not the scope of this task. I don't think that we should add `boolean` prop `defaultTime`(the name does not indicate the type of this prop). Add task in jira to refactor timepicker, don't do this in this task.

monikaracka (approved):

PR #705, opened 2023-10-26
Ds 919 missing aria attributes

zuzannanap:

- adds activeItemId and setActiveItemId props to breadcrumb
- changes onClick to onRateClick prop in Rating

## Description:

it also adds aria-attributes: aria-current, aria-pressed and aria-expanded to components


packages/design-system/src/components/Breadcrumbs/index.tsx
```diff
@@ -70,87 +72,97 @@ const Breadcrumbs: FC<BreadcrumbsProps> = ({
     icon: separatorIcon,
   };
 
+  console.log(activeItemId);
```
monikaracka, line 75: to remove

monikaracka (approved):

PR #713, opened 2023-10-30
fix/DS-911 - datepicker

monikaracka:
## Description:
- fix typing in input 
- remove the double labels and supporting texts


pauzim (approved):
Works great 🙏 One thing I think was lost somewhere during refactors is last day of range styles - now it looks like it's invisible 🤔

image


Also I can't remember if the label should be red or not?
image

Things to consider (for separate task) - currently it looks like we don't have validation for the range and it looks like this: 
image

PR #728, opened 2023-11-13
Mobile/avatar

kielbasa-elp:
## Description:


monikaracka (approved):
As we discussed, types should be in a separate file. Apart from that, it looks good to me 👏

PR #744, opened 2023-11-20
Mobile/checkbox

kielbasa-elp:
## Description:


monikaracka (approved):
I'm considering if we could use `Label` and `SupportingText` components here, but you used context and custom text components, which, I think, are better for this.

PR #752, opened 2023-12-01
Enable clickable input prefix area

pauzim:
## Description:

Make input prefix area clickable


olborska-elp (approved):
I can click area around icon 👌

PR #753, opened 2023-12-01
DS-959: Enable custom styles for dropdown options

pauzim:
## Description:

-> Optional dropdown option param with custom styles for option content, icon and label text.


olborska-elp (approved):
I can customize color of option. When I merged #752 with `DatePicker` fixes, the tests passed.

PR #761, opened 2023-12-05
DS-967: Revert `useSidebar` and `SidebarContext`, fix stories

pauzim:
## Description:
- Add `useSidebar` and `SidebarContext`
- Removed props `isCollapsed`, `toggleCollapse`
- Changed `collapseButton` type: boolean -> ICollapseButton


olborska-elp (approved):
Here we go again 🎉

PR #764, opened 2023-12-08
[POC] DS-917 fix sorting table

olborska-elp:
## Description:

Fix sorting table content for non-primitive data.
breaking-changes: add generic type for `TableProps`


pauzim (approved):
Good job! 🙌

PR #765, opened 2023-12-11
DS-970: Use `dateFormat` prop as dateFormat also in month

pauzim:
## Description:

Enable use dateFormat also in month picker variant


olborska-elp (approved):
Correct date for month and month/year option. Should we show the same placeholder as date format by default?

PR #766, opened 2023-12-11
DS-973: Select Dropdown: Black color for selected item

pauzim:
## Description:


olborska-elp (approved):
Black is black

PR #769, opened 2023-12-12
DS-976: Enable passing custom label with `formatOptionLabel`

pauzim:
## Description:

Enabled usage of `formatOptionLabel`


pauzim:
> It works as expected. In the next PR we can check if it is possible to add storybook args for it.

Done already 😁

PR #788, opened 2023-12-20
DS-969 - datepicker header in month variant

monikaracka:
## Description:

- in `month` variant in Datepicker, if the month picker is active, only year is displayed in the header
- in `month` variant in Datepicker if year picker is active, year's period is displayed


jakmarch (approved):
LGTM 👍

